PR #2, opened 2020-06-26
Support uploading results to BigQuery

codeblooded:
This commit adds support for mounting a secret volume, allowing driver pods to use a GCP service account. Developers should use the service account to write benchmark results and artifacts to non-volatile storage.

The developer must manually add a Kubernetes secret with a security key for the service account. Then, they specify the name of the secret in an environment variable: $GCP_KEY_SECRET. Finally, the system will detect the presence of the environment variable and attempt to mount the secret into the driver component's container.

The previous version of the orch package introduced a SpecBuilder type. Executors used this type to create Pod instances. This type allowed the Pod spec to be broken into multiple methods. Each method focused on one aspect of the pod, such as environment variables. Unfortunately, volumes requiring sharing data across these methods. Therefore, this commit replaces the SpecBuilder type with one function `makePod` and its helper `scenarioJSON`.

Finally, this commit intentionally lacks automated testing of the new functions and secret business logic. The team plans to move to a custom controller, causing this logic to be split into a new package. During the split, tests will be added using the ginkgo framework.


codeblooded:
Noticed naming issue, so marking as draft until resolved. Will squash commits when merging.

codeblooded:
This currently works by selecting the result table name when building the Docker image for the driver. The service proto definition has a field for this table name, but I'm not sure if we should add this business logic there. It could just as easily be an environment variable on the client (like the $GCP_KEY_SECRET) or part of the Custom Resource Definition when we move to a custom controller. See internal bug for full discussion.

benchmarks/k8s/images/driver/Dockerfile
```diff
@@ -12,45 +12,70 @@
...
+ARG GITREF=master
+
+RUN apt-get update && apt-get install -y git
+
+RUN mkdir -p /grpc/grpc
+WORKDIR /grpc/grpc
+
+RUN git clone https://github.com/$REPOSITORY.git .
+RUN git submodule update --init
+FROM l.gcr.io/google/bazel:0.17.1
```
codeblooded, line 27: This multi-stage process addresses some feedback on the slow build times experienced by prototype users. It is much faster and appears to condense the size of the image by about 70%.

PR #3, opened 2020-06-29
Create GitHub Actions workflow for Go

codeblooded:
This commit adds a GitHub action as CI for test-infra.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,56 @@
+name: Go
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.14
+      id: go
+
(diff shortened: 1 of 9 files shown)


codeblooded:
Sure thing, @jtattermusch. I installed it on my fork. It detected some format issues, so I'll fix them and rebase to accurately test.

I also thought about splitting build, lint/format, vet and test into separate jobs that share the build cache. The last 3 would run in parallel after the build completes and it would report them as separate statuses in the PR. This may more quickly convey the problems and speed things up.

Also, should I pin goimports or golint to specific versions? They are at HEAD here.

jtattermusch:
> Sure thing, @jtattermusch. I installed it on my fork. It detected some format issues, so I'll fix them and rebase to accurately test.
> 
> I also thought about splitting build, lint/format, vet and test into separate jobs that share the build cache. The last 3 would run in parallel after the build completes and it would report them as separate statuses in the PR. This may more quickly convey the problems and speed things up.

We can start with the current form and adjust later as needed.
I'm not worried about test latency at this point as it looks like things will be fast enough at least for a start (and we can always improve)

> 
> Also, should I pin goimports or golint to specific versions? They are at HEAD here.

Can come as a future improvement.

PR #4, opened 2020-07-01
Update service for Kubernetes client-go v0.18.x

codeblooded:
The Go team recommends using a context.Context as the first parameter to provide timeouts, deadlines and cancellation functions.

The previous versions of the Kubernetes libraries did not use contexts. The new versions (v0.18+) include contexts, introducing a backwards incompatibility.

This commit updates to v0.18.02. It changes the versions in go.mod and adds contexts to the orch package. This will make it easier to use kubebuilder to create a custom controller.

⚠️ This change may not be "supported" by stable versions of GKE. I have not yet noticed any issues and reverting is easy.


codeblooded:
Closing since kubebuilder is working with v17. Saving this commit for when v18 hits us in the next months/year.

PR #5, opened 2020-07-06
Add kubebuilder, custom resource and controller

codeblooded:
<table>
<tr>
<td>
⚠️
</td>
<td>
<b>DO NOT MERGE!</b>
</td>
</tr>
</table>

This change adds the kubebuilder framework, a boat-load of dependencies (hundreds), the ginkgo test framework and -- most importantly -- a custom resource for building a controller.  This allows us to begin discussing what is happening, but it will likely not be merged AS-IS.


codeblooded:
This does not currently build, b/c GitHub actions is not configured to setup a virtual environment for integration testing (as scaffolded by kubebuilder). It does not have access to the cluster. To fix, we will also likely need to add Kind and some other service dependencies to our GitHub actions workflow.

api/v1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+package v1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "e2etest.grpc.io.e2etest.grpc.io", Version: "v1"}
```
codeblooded, line 29: This can't be right.

```suggestion
	GroupVersion = schema.GroupVersion{Group: "e2etest.grpc.io", Version: "v1"}
```

codeblooded:
When reviewing [loadtest_types](api/v1/loadtest_types.go), it is important to consider the [Kubernetes API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md). One convention is using pointers on primitives to signal optionality.

codeblooded:
Going to split this into multiple PRs. Closing this large one.

codeblooded:
Split so far into #10 and #11. Will continue as makes sense.

PR #12, opened 2020-07-23
Add defaults for LoadTests

codeblooded:
This change adds defaults for a LoadTest resource, defined in the defaults package. These defaults are meant to customize the behavior of the system. They are not part of the controllers package, because these defaults may be shared with an admission or mutating webhook at some point.

The defaults package is placed under a pkg/ directory, a standard for signaling that a package is a public API. This conveys that the defaults are intended to be modified. It does not provide any guarantees about backwards compatibility.

In addition, this adds the package [github.com/pkg/errors](https://github.com/pkg/errors) to provide better wrapping and unwrapping for errors.


codeblooded:
@paulosjca This PR has been updated to use a config file as we discussed. I chose YAML since we use it for everything else. I also was able to use the YAML parser that is used by Kubernetes (and already a transitive dependency). I made it a direct dependency. This has an example config file: [example_config.go](https://github.com/grpc/test-infra/pull/12/files#diff-184a25755a707bb015b820e37d2608b2). I need to add much more documentation, so I plan to do that in another PR.

codeblooded:
@paulosjca I added documentation comments to the internal `imageMap` structure, constructor and methods as discussed. I did so in a new commit, since the review is in progress. 👍🏻

codeblooded:
@jtattermusch understood for now. we no longer hardcode any container images. We require a config file which specifies them. After discussion with @paulosjca, we decided this file should not necessarily be checked into the repo. The remaining references to the GitHub package registry are example config files and unit tests.

codeblooded:
The user (external or internal) which deploys to a cluster will need to supply their own config.

PR #13, opened 2020-07-28
Fix lint and fmt checks in GitHub actions

codeblooded:
The exit codes for the fmt and golint checks were not being reflected in the status of a build. In addition, problems were not being properly listed. This change fixes both issues by wrapping with a `test -z` command and echoing the contents.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -43,8 +43,11 @@ jobs:
 
     - name: Check for format and lint issues
       run: |
-        test -z $(gofmt -s -d -l .) # see https://github.com/golang/go/issues/24230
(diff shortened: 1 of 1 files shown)


codeblooded:
This is **Failing** as expected, since the linter and fmt checks did not work on some prior PRs. We can merge alongside fixes for style issues.

codeblooded:
I think the best thing to do is merge this and follow with another PR to fix the actual lint issues. @paulosjca @jtattermusch

codeblooded:
Note: our repository settings do not require all passing checks to merge. I'm pretty sure you can require them. But, I guess this is desirable for situations like this.

PR #14, opened 2020-07-28
Fix lint issues

codeblooded:
This branch is based off of #13, which should be merged first. This PR shows the actions working as opposed to the failing lint checks in #13.

### DIFF
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -37,6 +37,9 @@ type LoadTestReconciler struct {
 // +kubebuilder:rbac:groups=e2etest.grpc.io,resources=loadtests,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=e2etest.grpc.io,resources=loadtests/status,verbs=get;update;patch
 
+// Reconcile attempts to bring the current state of the loadtest into
+// agreement with its declared spec. This may mean provisioning
+// resources, doing nothing or handling the termination of its pods.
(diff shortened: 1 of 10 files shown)


controllers/loadtest_controller.go
```diff
@@ -37,6 +37,9 @@ type LoadTestReconciler struct {
 // +kubebuilder:rbac:groups=e2etest.grpc.io,resources=loadtests,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=e2etest.grpc.io,resources=loadtests/status,verbs=get;update;patch
 
+// Reconcile attempts to bring the current state of the loadtest into
```
paulosjca, line 40: I would write "load test" rather than "loadtest". In previous projects we settled to use "load_test" in file names etc. as well, but that is secondary.

codeblooded: I think that makes sense for comments. In kubernetes, however, it's confusing because it is a single entity. So you'd run `kubectl describe loadtest foo`. In code, I refer to it as `loadtest` in order to denote that it's a singular resource. This is how kubebuilder generates things. For example, the file `loadtest_types.go`. I thought about `loadTest` but swayed away from it. Is it better to separate it or denote it as `LoadTest` (the resource) in the comment? @paulosjca

paulosjca: I would write "load test" in the comment (since it is not code).

codeblooded: I couldn't agree more now. That's a good point. I think it is much more readable. Thanks, @paulosjca 😃

PR #15, opened 2020-08-03
Add pod creation logic

codeblooded:
This PR adds tests and methods to consistently define pod specs and metadata for load test components.

### DIFF
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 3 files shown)


codeblooded:
Adding @wanlin31 as reviewer, so she can reference this for her missing pod work. Also added @paulosjca so we can begin discussions.

codeblooded:
I still need to add volume mounts for GCP secrets (allowing saves to BigQuery) and logic to handle scenarios. These will come as separate PRs, since they are likely to cross multiple packages.

PR #16, opened 2020-08-03
Add ruby image

wanlin31:
This change adds a Dockerfile for ruby in the containers/runtime/ruby
directory.

This image can successfully start a ruby worker from grpc/grpc.


containers/runtime/ruby/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+RUN mkdir -p /src/workspace
+WORKDIR /src/workspace
+
+RUN apt-get update && apt-get install -y \
+  git\
+  time && \
+  apt-get clean
+
+CMD ["bash"]
```
codeblooded, line 24: GitHub is complaining b/c this file does not end in 1 blank new line. I think that's a linux convention. I'd add it just to shut it up. Otherwise, LGTM.

```suggestion
CMD ["bash"]

```

containers/runtime/ruby/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+FROM ruby:2.5
+
+RUN mkdir -p /src/workspace
+WORKDIR /src/workspace
+
+RUN apt-get update && apt-get install -y \
+  git\
```
paulosjca, line 20: I would add a space before \ for consistency. Other than that, LGTM.

PR #17, opened 2020-08-04
Fix code generation comment

codeblooded:
Whenever the api/v1/loadtest_types.go file is changed, the developer must regenerate the YAML files. Previously, this comment did not mention rebuilding the manifests. This change rewords it to include both the `make all` and `make manifests` targets.

### DIFF
--- a/api/v1/loadtest_types.go
+++ b/api/v1/loadtest_types.go
@@ -21,7 +21,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// NOTE: AFTER EDITS, YOU MUST RUN `make` TO REGENERATE CODE.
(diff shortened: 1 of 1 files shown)


api/v1/loadtest_types.go
```diff
@@ -21,7 +21,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// NOTE: AFTER EDITS, YOU MUST RUN `make` TO REGENERATE CODE.
+// NOTE: CODE IS GENERATED FROM THIS FILE!
```
paulosjca, line 24: I would keep the comment more or less as before, except for `make manifests` AND `make` insted of just `make`.

codeblooded: Should be fixed. What do you think?

PR #19, opened 2020-08-10
Generate unique component names when unspecified

codeblooded:
Each driver, server and client component must have a relatively unique name. This means their name cannot be identical to another component with the same role in the same load test. Previously, this name was manually provided.

This change generates unique names for components where names are not manually set. To ensure relative uniqueness, it uses Google's UUID library to generate a token.

### DIFF
--- a/pkg/defaults/loadtest_defaults.go
+++ b/pkg/defaults/loadtest_defaults.go
@@ -1,6 +1,7 @@
 package defaults
 
 import (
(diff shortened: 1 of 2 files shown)


codeblooded:
Fixed format issue in check. The other issue is a flake outside of our control (kubernetes-related, I believe). I'm re-running the test now, but I think it should pass.

PR #21, opened 2020-08-11
Mount volumes with scenarios

codeblooded:
This change creates a volume with a ConfigMap containing the Scenario as the source. This volume is then mounted into a driver pod to supply the configuration for the test.

### DIFF
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


controllers/loadtest_controller.go
```diff
@@ -153,7 +198,18 @@ func newDriverPod(loadtest *grpcv1.LoadTest, component *grpcv1.Component) (*core
 		return nil, err
 	}
 
-	addDriverPort(&pod.Spec.Containers[0])
+	spec := &pod.Spec
+
+	// TODO: Avoid referencing containers by index, use names
+	addDriverPort(&spec.Containers[0])
```
codeblooded, line 204: Will follow up with another PR that contains function for looping and finding the appropriate container. This will be useful in multiple places in the future.

controllers/loadtest_controller.go
```diff
@@ -153,7 +198,18 @@ func newDriverPod(loadtest *grpcv1.LoadTest, component *grpcv1.Component) (*core
...
+
+	// TODO: Handle more than 1 scenario
+
+	// At this point, we assume validation will have caught any loadtest
+	// without a scenario.
```
codeblooded, line 209: We need to add this logic to the controller itself. I don't rely on returning and propagating an error here, because we should not wait until creating the driver to see there is no scenario on the load test. We should mark it as an error on the very first reconcile before creating anything.

PR #22, opened 2020-08-11
Build driver code in driver Dockerfile

codeblooded:
The previous version of the driver image prepared an environment for running the driver but did not include a compiled version of the driver in the container image. The driver code would be built at runtime. However, this would result in slow Java and Go tests, since the driver takes much longer to build than Java or Go's workers.

This change compiles the latest driver from the master branch of grpc/grpc whenever the container image is rebuilt. This allows a driver to start almost instantly!

In addition, it uses the `--scenario_file` flag to accept a JSON file as an argument. This *differs* from the prototype, because it moves towards mounted volumes over environment variables. Environment variables are problematic when the JSON is not minified. Whitespace caused the driver to parse the JSON as separate arguments. In addition, there could be limits to the maximum characters in an environment variable or a single command. Volumes allow kubernetes to mount a file with the JSON that the driver can parse, circumventing these issues. It is also compatible with #21.

### DIFF
--- a/containers/runtime/driver/Dockerfile
+++ b/containers/runtime/driver/Dockerfile
@@ -14,8 +14,29 @@
 
 FROM debian:buster
 
-RUN mkdir -p /src/workspace
+ARG REPOSITORY=grpc/grpc
+ARG GITREF=master
+
+RUN apt-get update && apt-get install -y git
+
+RUN mkdir -p /grpc/grpc
+WORKDIR /grpc/grpc
+
+RUN git clone https://github.com/$REPOSITORY.git .
+RUN git submodule update --init
+
+FROM l.gcr.io/google/bazel:0.17.1
+
+COPY --from=0 /grpc/grpc /src/workspace
+RUN mkdir -p /tmp/build_output
 WORKDIR /src/workspace
+RUN bazel --output_user_root=/tmp/build_output build //test/cpp/qps:qps_json_driver
+
+FROM debian:buster
+
+WORKDIR /grpc/grpc
+COPY --from=1 /tmp/build_output /tmp/build_output
+COPY --from=1 /src/workspace .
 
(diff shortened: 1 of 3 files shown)


codeblooded:
@grpc/test-infra I know volumes are a change from what was planned, but I think it is the best approach. The minified JSON issue was a pain, and volumes prepare the way for multiple scenarios in the future.

containers/runtime/driver/Dockerfile
```diff
@@ -14,8 +14,29 @@
...
+RUN git submodule update --init
+
+FROM l.gcr.io/google/bazel:0.17.1
+
+COPY --from=0 /grpc/grpc /src/workspace
+RUN mkdir -p /tmp/build_output
 WORKDIR /src/workspace
+RUN bazel --output_user_root=/tmp/build_output build //test/cpp/qps:qps_json_driver
+
+FROM debian:buster
```
gnossen, line 35: Any particular reasoning behind `debian:buster`?

codeblooded: @jtattermusch said we should update from stretch to latest debian stable. I think that's buster. Is there another base image that would be good here?

gnossen: If you really wanted to be lean and mean, you could build on top of `scratch`. That's literally a blank filesystem. For (a ton) more effort, you could get a (somewhat) slimmer image. This is more difficult for languages like Python, which are Gordian knot of dependencies, but pretty manageable for always-statically-linked language like Go. Kubernetes-associated projects tend to use this approach since their images are used so frequently.

It's probably not worth implementing, but I figured it was worth mentioning.

codeblooded: Good to note! Thanks 👍🏻

PR #23, opened 2020-08-13
Add missing pods logic

wanlin31:


controllers/loadtest_controller.go
```diff
@@ -125,6 +125,70 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	return ctrl.Result{}, nil
 }
 
+// CheckMissingPods attempts to check if any pods is missing for current
+// LoadTest.
+func CheckMissingPods(currentLoadTest *grpcv1.LoadTest, allRunningPods *corev1.PodList) []*grpcv1.Component {
+
+	//Creat maps for Clients and Servers for current currentLoadTest
```
paulosjca, line 132: Creat -> Create. "for current currentLoadTest" is probably not what you want to say.

controllers/loadtest_controller_test.go
```diff
@@ -502,3 +501,134 @@ var _ = Describe("Pod Creation", func() {
...
+			It("return the full list", func() {
+				Expect(checkIfEqual(returnedList, expectedReturnList)).To(Equal(true))
+			})
+		})
+	})
+
+	Describe("some of pods from current loadtest is running", func() {
+
+		BeforeEach(func() {
+			//Add pod with defaults.ComponentNameLabel: server-1, pod with defaults.ComponentNameLabel: client-2, driver
```
paulosjca, line 553: This comment could be clearer.

controllers/loadtest_controller.go
```diff
@@ -125,6 +125,70 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	for _, eachPod := range allRunningPods.Items {
+		loadTestLabelForEachPod := eachPod.Labels[defaults.LoadTestLabel]
+		roleLabelForEachPod := eachPod.Labels[defaults.RoleLabel]
+		componentNameLabelForEachPod := eachPod.Labels[defaults.ComponentNameLabel]
+
+		// ignore pods don't belong to current loadtest
+		if loadTestLabelForEachPod != currentLoadTest.Name {
+			continue
+		}
+		if roleLabelForEachPod == "driver" {
```
codeblooded, line 157: Lets not rely on string literals in-case we change the component role at some point. 

```suggestion
		if roleLabelForEachPod == defaults.DriverRole {
```

controllers/loadtest_controller.go
```diff
@@ -125,6 +125,70 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			// check if it is a driver
+			if *currentLoadTest.Spec.Driver.Component.Name == componentNameLabelForEachPod {
+				foundDriver = true
+			}
+		} else if roleLabelForEachPod == "client" {
```
codeblooded, line 162:
```suggestion
		} else if roleLabelForEachPod == defaults.ClientRole {
```

controllers/loadtest_controller.go
```diff
@@ -125,6 +125,70 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			// check if it is a client
+			if _, ok := CurrentLoadTestClientPodMap[componentNameLabelForEachPod]; ok {
+				delete(CurrentLoadTestClientPodMap, componentNameLabelForEachPod)
+			}
+		} else if roleLabelForEachPod == "server" {
```
codeblooded, line 167:
```suggestion
		} else if roleLabelForEachPod == defaults.ServerRole {
```

controllers/loadtest_controller_test.go
```diff
@@ -502,3 +501,134 @@ var _ = Describe("Pod Creation", func() {
...
+
+	JustBeforeEach(func() {
+		returnedList = CheckMissingPods(currentLoadTest, allRunningPods)
+	})
+
+	AfterEach(func() {
+		expectedReturnList = []*grpcv1.Component{}
+		newThing := corev1.PodList{Items: []corev1.Pod{}}
+		allRunningPods = &newThing
+	})
```
codeblooded, line 520: The slice declaration in line 508 is fine, because this is a nil slice. It is not actually initialized until you `amend` to it. I would prefer if the assignment in lines 507 and 509 are moved into a `BeforeEach`. I see your logic and its valid for the current state of the code. But, imagine in the future that Paulo or I come add an `It` block. We modify `currentLoadTest`. Then, we're not hermetic. It is better to put all initializations in a `BeforeEach`.

Prefer a fresh initialization with a `BeforeEach` over a clean-up with `AfterEach`. Clean-ups are much harder to get right, because each test must rely on the successful clean-up of the previous one.

Other than this, this section LGTM.

controllers/loadtest_controller_test.go
```diff
@@ -502,3 +501,134 @@ var _ = Describe("Pod Creation", func() {
...
+			for i := 0; i < len(currentLoadTest.Spec.Servers); i++ {
+				expectedReturnList = append(expectedReturnList, &currentLoadTest.Spec.Servers[i].Component)
+			}
+			expectedReturnList = append(expectedReturnList, &currentLoadTest.Spec.Driver.Component)
+		})
+
+		Context("allRunningPods is empty", func() {
+			allRunningPods = &corev1.PodList{Items: []corev1.Pod{}}
+			It("return the full list", func() {
+				Expect(checkIfEqual(returnedList, expectedReturnList)).To(Equal(true))
```
codeblooded, line 537: The `checkIfEqual` helper is okay for now. Let's prefer the `BeTrue()` matchers, like: 

```suggestion
				Expect(checkIfEqual(returnedList, expectedReturnList)).To(BeTrue())
```

fun experiment: If you really wanted to, you could create your own matcher, like `Expect(returnedList).To(ContainEquivalentNamesToList(expectedReturnList))`. See [GoMega custom matchers](https://onsi.github.io/gomega/#adding-your-own-matchers). I have never done this.

controllers/loadtest_controller_test.go
```diff
@@ -502,3 +501,134 @@ var _ = Describe("Pod Creation", func() {
...
+
+	Describe("some of pods from current loadtest is running", func() {
+
+		BeforeEach(func() {
+			//Add pod with defaults.ComponentNameLabel: server-1, pod with defaults.ComponentNameLabel: client-2, driver
+			//to running podlist
+			allRunningPods.Items = append(allRunningPods.Items,
+				corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "random_name",
```
codeblooded, line 558: Great example. Just an option to teach something with Kubernetes. Names must be valid subdomains. So, underscores are not okay, but dashes are. This is fine for unit tests. But, if you want it to match the actual pods we'll see, switch out the underscores for dashes. 👍🏻

controllers/loadtest_controller.go
wanlin31: I will definitely check all my spelling more carefully!

controllers/loadtest_controller_test.go
wanlin31: Agreed for the concept of getting a fresh start other than have next test depend on previous clean up. Re-construct the test structure to reflect the change.

controllers/loadtest_controller_test.go
wanlin31: Great! I definitely should look more into the provided matchers!

controllers/loadtest_controller_test.go
wanlin31: Will do! Thanks for teaching me here and there. :)

controllers/loadtest_controller_test.go
wanlin31: I was wondering if I should delete that comment all together, it is self explanatory when look at the codes.

controllers/loadtest_controller.go
```diff
@@ -125,6 +126,67 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	return ctrl.Result{}, nil
 }
 
+// CheckMissingPods attempts to check if any pods is missing for current
+// LoadTest.
+func CheckMissingPods(currentLoadTest *grpcv1.LoadTest, allRunningPods *corev1.PodList) []*grpcv1.Component {
+
+	//Creat maps for Clients and Servers for current currentLoadTest
+	CurrentLoadTestClientPodMap := make(map[string]*grpcv1.Component)
```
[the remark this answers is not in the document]
wanlin31: Yes changed.

controllers/loadtest_controller.go
wanlin31: Agreed! Should minimize hardcoded names or labels.

api/v1/loadtest_types.go
```diff
@@ -208,6 +208,24 @@ type LoadTestSpec struct {
...
+	// Servers are a list of components that receive traffic from
+	// clients.
+	// +optional
+	Servers []Server `json:"servers,omitempty"`
+
+	// Clients are a list of components that send traffic to servers.
+	// +optional
+	Clients []Client `json:"clients,omitempty"`
+}
+
```
codeblooded, line 228: This is a great struct. I think we should consider moving this out of api/v1/loadtest_types.go for 1 reason. Everything in api/ is considered public API, and Kubebuilder generates a lot of code for it. For instance, it creates `DeepCopy` methods. Since this type is only meant for internal use, I think it would be better to move it into controllers/loadtest_controller.go as an unexported type. Then, we can remove the `//+optional` markers and no code will be generated for it. If we leave it here, we will need to generate that code as part of this PR. What are your thoughts?

controllers/loadtest_controller.go
```diff
@@ -125,6 +124,69 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	return ctrl.Result{}, nil
 }
 
+// checkMissingPods attempts to check if any pods is missing for current
+// loadtest.
```
codeblooded, line 128: nit: @paulosjca may chime in, since he's good with documentation... but I would consider elaborating a little more about what it accepts and what is returned. Otherwise, good to me 👍🏻

paulosjca: Just a small thing, when using 'load test' as plain English, I would use two words (this may be true elsewhere as well). Also, 'the current load test'.

api/v1/loadtest_types.go
wanlin31: I agree, I don't think LoadTestMissing will be used externally, I put it there just because I though all the types should go to that file, turned out to be not. :) I will move it to loadtest_controller.go, location-wise, is there a fashion to follow? I saw you put the LoadTestReconciler right on top, should my struct be closer to my func or it should be declared on the top just after LoadTestReconciler struct?

controllers/loadtest_controller.go
wanlin31: I will modify the loadtest to load test, that is a great suggestion! Also try to explain the method clearer.

api/v1/loadtest_types.go
codeblooded: If you do move it, be sure to run `git restore --source=upstream/master api/v1/zz_generated.deepcopy.go` to restore the previous version of the generated code.

wanlin31: Also do I still need to have the comments in there?

codeblooded: @wanlin31 I like to keep things close to where they are used. Since it's not exported, I would not put it above the LoadTestReconciler. -- Be sure to run `git restore --source upstream/master api/v1/zz_generated.deepcopy.go`

codeblooded: @wanlin31 you do not need the `// +optional` markers, but I'd leave the other comments. @paulosjca had a discussion about this recently with my pkg/defaults code. Even though comments are not required for unexported types, it can help other people understand it. It doesn't ever hurt to leave it :)

wanlin31: :) Absolutely, I add a little more explanation there just to help people understand what LoadTestMissing is.

controllers/loadtest_controller.go
```diff
@@ -124,11 +141,13 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	return ctrl.Result{}, nil
 }
 
-// checkMissingPods attempts to check if any pods is missing for current
-// loadtest.
-func checkMissingPods(currentLoadTest *grpcv1.LoadTest, allRunningPods *corev1.PodList) *grpcv1.LoadTestMissing {
+// checkMissingPods attempts to check if any required component is missing from
+// the current load test. It takes reference of the current load test and a pod
+// list that contains all running pods at the moment returns all missing
```
codeblooded, line 146: nit: this sentence is a little complex, perhaps adding a command and switching to a gerund may make it a little more readable.

```suggestion
// list that contains all running pods at the moment, returning all missing
```

PR #27, opened 2020-08-19
Add ready init container

codeblooded:
The driver needs to wait for all worker pods to be assigned an IP address and be ready before running. This change introduces the ready init container. This container allows the driver to block until a pods with matching labels become ready or a timeout is reached.

### DIFF
--- a/containers/init/ready/Dockerfile
+++ b/containers/init/ready/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2020 gRPC authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


codeblooded:
@wanlin31 As you can see, I reused some of your ideas with an algorithm that deletes from a set when matches are found. Your work made mine better :)

codeblooded:
@grpc/test-infra note code coverage is not great on this, because a test environment is hard to setup. This executable relies on a Kubernetes config + server, a set of environment variables and may block. I hope to add better tests in the future. Since the main matching logic is tested, I think this is decent for an MVP.

codeblooded:
I need to handle port numbers in the output list that this writes to a file, but I'll figure out where to place those in a separate PR.

PR #29, opened 2020-08-21
Set default namespace on load tests and their pods

codeblooded:
The previous version of the code did not specify a namespace when creating pods for load tests. This led to a panic, because the default namespace is not inferred.

This change explicitly sets the namespace of pods to match the namespace of the load test. Though unlikely to be required, it also ensures that the load test lives in the default namespace if another has not been specified.

### DIFF
--- a/pkg/defaults/loadtest_defaults.go
+++ b/pkg/defaults/loadtest_defaults.go
@@ -4,6 +4,8 @@ import (
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
 
+	corev1 "k8s.io/api/core/v1"
+
 	grpcv1 "github.com/grpc/test-infra/api/v1"
 )
 
@@ -17,6 +19,10 @@ func CopyWithDefaults(d *Defaults, loadtest *grpcv1.LoadTest) (*grpcv1.LoadTest,
 	im := newImageMap(d.Languages)
 	test := loadtest.DeepCopy()
 	spec := &test.Spec
+
+	if test.Namespace == "" {
+		test.Namespace = corev1.NamespaceDefault
+	}
 
 	if spec.Driver == nil {
 		spec.Driver = &grpcv1.Driver{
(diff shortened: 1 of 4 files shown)


pkg/defaults/loadtest_defaults.go
```diff
@@ -18,6 +20,10 @@ func CopyWithDefaults(d *Defaults, loadtest *grpcv1.LoadTest) (*grpcv1.LoadTest,
 	test := loadtest.DeepCopy()
 	spec := &test.Spec
 
+	if test.Namespace == "" {
+		test.Namespace = corev1.NamespaceDefault
+	}
+
```
wanlin31, line 26: Do we encourage users to set up the load test Namespace? I think if we don't have too many deployment and made sure they have different names, we would not necessary need to logically group our stuff, so it would be okay to have everything who's Namespace not specified live in the default Namespace. I asked because Namespace was not mentioned in your example yaml file, just wondering what is our recommendation here (https://docs.google.com/document/d/15iuHZtZr6rucy9diqk91MCqyQxApkff69Rm2ANHFtw0/edit#heading=h.bl834lldomuy).

codeblooded: @wanlin31 Namespaces are used to group kubernetes resources. Our controller should work within the scope of any namespace. We do not have a namespace plan for an MVP. The default namespace is sufficient. For a future version, we may decide to use a namespace per user, per team, per language, to isolate CI from developers manually running tests, to isolate experimental changes from stable ones, etc. The downside to namespaces is that they can complicate scheduling logic.

wanlin31: Thanks for explanation. Everything looks good to me. :) A minor question about how to run load test, my understanding was that for MVP we will run manually but in the future the test would be scheduled automatically, is this correct?

codeblooded: @wanlin31 yes. See files from [698e3cf9fd](https://github.com/codeblooded/grpc-test-infra/tree/698e3cf9fd0bb26860ddcaf6ab72695a8f09aa5c/config/samples) as an example. The load test would be started using `kubectl apply -f e2etest.grpc.io_v1_loadtest.yaml`. When using `kubectl`, it sets the default namespace unless you tell it otherwise :)

wanlin31: Learned a lot! Thank you. :)

PR #30, opened 2020-08-21
Fix wrong tag name in defaults

codeblooded:
The LanguageDefault type in the defaults package used to have a `Name` field. This represented the name of the programming language and was renamed to `Language`. During the rename, the JSON tag was not changed. This commit updates the JSON tag.

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -47,7 +47,7 @@ type LanguageDefault struct {
 	// Language uniquely identifies a programming language. When the
 	// system encounters this name, it will select the build image and
 	// run image as the defaults.
-	Language string `json:"name"`
+	Language string `json:"language"`
 
 	// BuildImage specifies the default container image for building or
 	// assembling an executable or bundle for a language. This image


pkg/defaults/defaults.go
```diff
@@ -47,7 +47,7 @@ type LanguageDefault struct {
 	// Language uniquely identifies a programming language. When the
 	// system encounters this name, it will select the build image and
 	// run image as the defaults.
-	Language string `json:"name"`
+	Language string `json:"language"`
```
wanlin31, line 50: This is for my own learning :) : If we used to serialize Language struct with tag name, who would be deserializing our data? Do they also change "name" to "language", or they now have "language" and we just have to fix our side.  Could you point the history of this change?

codeblooded: @wanlin31 This is essentially the "schema" for our config file. If it were in use outside of us, it would be a backwards-incompatible change. However, the only current use is in [config/example_config.yaml](https://github.com/grpc/test-infra/blob/a2b80ecd861f543fb0b5f42783ef9a085743653b/config/example_config.yaml). It's always reasonable to ask to point to an old change. In this case, see a192e3c4dd0ba85cd0eaf19da097f86a208d27c2 and 4c1484df7287ba091dca959a240ec15dc87d221d.

wanlin31: Great explanation! I didn't realize that yaml.Unmarshal could parse YAML to JSON and use JSON to unmarshal into object, this made a lot more sense to me now! 👍

codeblooded: @wanlin31 Normally it doesn't. Kubernetes uses a special YAML package that works with the JSON tags: https://github.com/kubernetes-sigs/yaml

PR #33, opened 2020-08-26
WIP Integrate missing pod and create pod logic

codeblooded:


controllers/loadtest_controller.go
```diff
@@ -87,35 +87,40 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	if err = r.List(ctx, pods, client.InNamespace(req.Namespace)); err != nil {
 		log.Error(err, "failed to list pods", "namespace", req.Namespace)
+
 		// attempt to requeue with exponential back-off
 		return ctrl.Result{Requeue: true}, err
 	}
 
-	var loadtests grpcv1.LoadTestList
-	if err = r.List(ctx, &loadtests); err != nil {
+	loadtests := new(grpcv1.LoadTestList)
```
wanlin31, line 106: Do we have these new function already or we are trying to create them later?

codeblooded: The `new` function is actually built-in for Go. It can create a pointer to any type with zero values for all fields. It's equivalent to `&grpcv1.LoadTestList{}`, and sometime it just looks a little cleaner to me. 🙂

codeblooded: It differs from `make`.

The `new` function is used for structs. It says (1) allocate the memory for this thing and (2) return a pointer to this memory.

The `make` function does additional work, because it's used on complex data types. For a map or channel, we use `make`. These require some complex data structures underneath them that need to be setup. This is also why we can't add to a nil map or channel, but we can append to a nil slice. A slice is a simple type and just grows as things are amended. A map must use a hash table for keys. A channel must use things to make concurrent access safe.

codeblooded:
Closing in favor of smaller, better PRs.

controllers/loadtest_controller.go
wanlin31: Thank you :)

PR #38, opened 2020-08-27
Set build and run working directory to workspace

codeblooded:
This commit sets the working directory for the build init container and run container to the /src/workspace volume mount.

This does not set the working directory on the init container, because it does not require a specific working directory.

### DIFF
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -434,11 +434,12 @@ func newBuildContainer(build *grpcv1.Build) corev1.Container {
 	}
 
 	return corev1.Container{
-		Name:    buildInitContainer,
-		Image:   *build.Image,
-		Command: build.Command,
-		Args:    build.Args,
-		Env:     build.Env,
+		Name:       buildInitContainer,
+		Image:      *build.Image,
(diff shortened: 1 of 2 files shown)


codeblooded:
This change includes #37, so it should be merged first and this should be rebased.

codeblooded:
Just added missing word in commit message that @wanlin31 noticed. Good to merge once checks passed 🚀

codeblooded:
@wanlin31 The clone container needs to obtain a copy of the code at a specific point in time and place it in the /src/workspace path. It does not have to consider this path its working directory. Our containers/init/clone actually already sets and uses /src/workspace as it's working directory. But, a clone image doesn't have to.

Picture a Dockerfile for a clone image which copies a script _copy.sh_. This script clones some code and writes it to some directory. The developer sets `CMD ["./copy.sh"]` in the Dockerfile. If we override it to /src/workspace, we may break the functionality. 

The clone image does not require a shared input. It only requires we write its output to a shared directory. It's working directory may be where the output is written, or it may not. It's optional.

Note: for our clone image, we do use /src/workspace as the working directory. It's just not required.

PR #39, opened 2020-08-28
Use defaults to set ports and port arguments

codeblooded:
This change uses the default ports to expose the ports on the pod and supply the port flags to the run container.


codeblooded:
@grpc/test-infra Fixed some typos... let me know what you think.

PR #42, opened 2020-08-29
Switch example load test to use Go

codeblooded:
The Go load tests build and start much quicker and they are easier to read. For this reason, this pull request switches the example load test to use Go workers. In addition, it adds an example Go scenario and updates the README.md file to reflect these changes.


wanlin31:
I will update the basic instruction with the change. :)

codeblooded:
@wanlin31 Great thanks. 🤓

PR #43, opened 2020-09-01
Setup the ready init container for a driver

codeblooded:
This change configures an init container that waits for workers to become ready before allowing the driver to start. This init container also writes the IP addresses and ports of available workers to a file. This change shares that file over a volume with the driver's run container.

### DIFF
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -45,6 +45,22 @@ const cloneInitContainer = "clone"
 // binary or other bundle required to run the tests.
 const buildInitContainer = "build"
 
+// readyInitContainer holds the name of the init container that blocks a driver
+// from running until all worker pods are ready.
+const readyInitContainer = "ready"
+
+// readyVolume is the name of the volume that permits sharing files between the
+// ready init container and the driver's run container.
+const readyVolume = "worker-addresses"
+
+// readyMountPath is the absolute path where the ready volume should be mounted
+// in both the ready init container and the driver's run container.
+const readyMountPath = "/var/data/qps_workers"
+
+// readyOutputFile is the name of the file where the ready init container should
+// write all IP addresses and port numbers for ready workers.
+const readyOutputFile = readyMountPath + "/addresses"
+
 // runContainer holds the name of the main container where the test is executed.
 const runContainer = "run"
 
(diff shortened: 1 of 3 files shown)


codeblooded:
This could probably use a few more tests, but I'm thinking I will restructure them soon. The code is covered, but just want it to be a little cleaner.

codeblooded:
@gnossen Thanks, yes... there was #27. I should have mentioned it. I'll amend the commit message to link the two together in the git log.

controllers/loadtest_controller.go
```diff
@@ -336,6 +352,37 @@ func newScenarioFileEnvVar(scenario string) corev1.EnvVar {
...
+// file.
+func addReadyInitContainer(defs *defaults.Defaults, loadtest *grpcv1.LoadTest, podspec *corev1.PodSpec, container *corev1.Container) {
+	if defs == nil || podspec == nil || container == nil {
+		return
+	}
+
+	readyCont := newReadyContainer(defs, loadtest)
+	podspec.InitContainers = append(podspec.InitContainers, readyCont)
+
+	container.Env = append(container.Env, corev1.EnvVar{
```
wanlin31, line 371: Is the container here the runtime container of the DriverPod's PodSpec? If so, podspec.container[0] would be the same with passing a new parameter(container here) to the function, right? Justing trying to understand the reason behind it. :)

codeblooded: Yes, the container is the run container. It may not always be the container at index 0. It only is with our current system. So, I want to add some utility functions for common operations like finding and returning the run container. See:

https://github.com/codeblooded/grpc-test-infra/blob/e4dbf6b6d43799a47c24443e305d78a1ed3c874d/controllers/loadtest_controller.go#L397

The other benefit is testing. This function needs tests. But, consider going to test it. If I want to test this function, I don't want to create a whole driver pod. I just want to see that the output of this function is expected. In fact, creating a driver pod may be counter-productive. This function shouldn't depend on something being set on the pod already. So, a test that uses a fully-fleshed driver pod may hide if it does.

A better set of tests would pass it a `new(corev1.Container)` and a `new(corev1.Pod)` and a tiny load test. Then, I would check... does my new container have the volume mount? Does my pod have the volume? Does my container have the environment variable? I'd be testing the specific functionality of this function with the most minimal test objects possible to show it's behavior is correct.

wanlin31: "It may not always be the container at index 0." --> Can we have multiple run time container? Theoretically Kubernetes definitely support this otherwise it would not provide containers as an array, but what occasion int our case we need multiple run time then? Could not figure out..... 
I do agree that have a name like testcontainer is far readable than container[0] 👍 
" I don't want to create a whole driver pod. " --> I think you were passing a podspec to the function also, that means you need to have the podspec regardlessly, if we don't want to create a podspec for this we probably want to pass podspec.initContainers and testcontainer separately so it doesn't depend on the pod creation.
I agree the tests part, it would be great the parameters could be tested independently.

wanlin31: I guess these would not matter much in terms of functionality, just trying to learn how we write better code. :)

codeblooded: > Can we have multiple run time container? Theoretically Kubernetes definitely support this otherwise it would not provide containers as an array, but what occasion int our case we need multiple run time then? Could not figure out.....

Yes, there is the concept of a container called a sidecar. Often, this is a proxy like [Envoy](https://www.envoyproxy.io/) or a metrics collecting container. We don't need to get side-tracked on these things for now. I can explain some of it to you when you have some time. Just know that these are things that @gnossen may use to improve tests. We also need to be careful to not have deep conversations about some internal things on GitHub. So, stick to internal chat and meet for these questions.

> I think you were passing a podspec to the function also, that means you need to have the podspec regardlessly, if we don't want to create a podspec for this we probably want to pass podspec.initContainers and testcontainer separately so it doesn't depend on the pod creation.

Correct, we do not need the `container` pointer if we did the lookup for the container on the PodSpec itself. I have chosen to not perform any lookup in this function. The caller can specifically tell me which Container inside of the PodSpec (through a pointer) to add the data to. This consolidates the lookup for the run container into a single place: the newDriverPod function.

wanlin31: Thank you for the explanation! Got it!

PR #44, opened 2020-09-01
Use less commonly used ports as defaults

codeblooded:
The default ports used by kubebuilder are commonly used by other services. For example, port :8080 is used by many web applications. This change switches to ports that are less commonly used.


codeblooded:
@paulosjca Amended. There are 2 ports, it's just awkward b/c one is a string: https://github.com/grpc/test-infra/pull/44/files#diff-c11f4f58679c6fc3508a32763a2b8d61R80-R81

PR #46, opened 2020-09-02
Fix deployment issues 

codeblooded:
This change fixes the issues with the Dockerfile, RBAC roles and defaults file to make deployments possible. In particular, it:

 - Switches the base image to `debian:buster` in the main Dockerfile. The distroless image had some issues with accessing the defaults file.

 - Adds CRUD permissions to pods and pod statuses for the manager, which encapsulates the controller.

 - Sets a default for the `-defaults-file` flag. This default points to the location of config/defaults.yaml. Similarly, the Dockerfile is updated to copy and adjust permissions on this file.

As a result of these changes, contributors must place a valid config file at the path config/defaults.yaml before building the docker image.

In addition, it adds the standard location for a customized config, config/defaults.yaml, to the .gitignore. This prevents developers from accidentally checking it in.

*With these changes, the controller has been deployed to the cluster.* :shipit:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@ bin/*
 *.orig
 .idea/
 
+# Hide internal configs
+config/defaults.yaml
+
 # Dependency directories (remove the comment below to include it)
 # vendor/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,18 +10,14 @@ COPY go.sum go.sum
 RUN go mod download
 
 # Copy the go source
-COPY main.go main.go
-COPY api/ api/
-COPY controllers/ controllers/
+COPY . .
+RUN chmod 777 /workspace/config/defaults.yaml
 
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
 
-# Use distroless as minimal base image to package the manager binary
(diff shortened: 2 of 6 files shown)


codeblooded:
@paulosjca 

> If someone wants to put the configuration in a non-standard location, how would the flag be passed in?

If someone wants to put the configuration in a non-standard location, they will need to devise their own way of deploying it to the cluster. You are free to pass the flag when you invoke the binary. -- When using kubebuilder deployments, the flag is often overridden to pass flags for its proxies. Most of the deployment YAML is generated, so it does not make sense to edit those files.

> In a related question, is the process of deploying the controller documented?

The process for deploying a cluster will be documented internally off GitHub as you requested for our specific setup. I plan to also document on GitHub how external users can deploy to their own clusters. This change fixes issues that made a deployment possible.

codeblooded:
@paulosjca To add to my reply, deploying this is fairly straightforward if the cluster has been setup properly. It is a matter of connecting and running the following:

```sh
export IMG=<path to your image on a registry>
make docker-build
make docker-push
make install deploy 
```

The fact that we are in a non-public GCR registry also complicates the process. It adds internal details to our generated YAML code, so a user has to discard those to not leak information. This would be much easier and cleaner if the registry was public.

paulosjca (approved):

PR #50, opened 2020-09-09
Add isolation through namespaces

codeblooded:
This commit makes it possible for multiple managers to run on the same cluster, considering resources only in a designated namespace. To do so, it introduces:

 - The `ComponentNamespace` option in the defaults. This is the default namespace for resources created by the controller.

 - The `-namespace` flag. This restricts the controller to only reconcile load tests and their pods in a specific namespace. When omited, all namespaces are considered.

---

**This change allows multiple developers to use managers all working on the same cluster.** To do so, a developer would create a new namespace for themselves. They would modify their config file (see config/example_config.yaml), setting the `componentNamespace` to be their new namespace. Finally, they would start the manager with the `-namespace=<their-namespace>` flag. This would restrict it to reconciling and creating resources in their namespace by default. -- This change should have no effect on the **prod** version.

### DIFF
--- a/config/example_config.yaml
+++ b/config/example_config.yaml
@@ -1,5 +1,7 @@
 # See ../pkg/defaults/defaults.go for documentation on each field.
 
+componentNamespace: default
+
 driverPool: drivers
 
 workerPool: workers-16core
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -113,11 +113,14 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Fetch the current state of the world.
 
-	nodes := new(corev1.NodeList)
-	if err = r.List(ctx, nodes); err != nil {
-		log.Error(err, "failed to list nodes")
-		return ctrl.Result{Requeue: true}, err
-	}
+	// Listing nodes will be required to achieve gang scheduling; however, this is
+	// not implemented at this time. Therefore, it is commented here for later:
+	//
+	//nodes := new(corev1.NodeList)
+	//if err = r.List(ctx, nodes); err != nil {
+	//	log.Error(err, "failed to list nodes")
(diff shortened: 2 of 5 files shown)


wanlin31:
Did this mean that the associated pods will be created in unique namespaces, therefor the anti affinity rule will not affect scheduling pods with different namespaces on the same node?

codeblooded:
@wanlin31 The anti-affinities will not likely be honored across namespaces. This is an open feature request, see  kubernetes/kubernetes#68827. For us, it should not be a problem. We are more interested in developing and debugging the process, not the quality of the metrics collected during development. In prod, they are all in 1 namespace, so it will work.

wanlin31:
> @wanlin31 The anti-affinities will not likely be honored across namespaces. This is an open feature request, see [kubernetes/kubernetes#68827](https://github.com/kubernetes/kubernetes/issues/68827). For us, it should not be a problem. We are more interested in developing and debugging the process, not the quality of the metrics collected during development. In prod, they are all in 1 namespace, so it will work.

Right, as you said anti-affinity would not be honored across namespace, and for current situation we have 4 nodes unless the previous load test owner clean up the space otherwise the next load test would not able to assign pods to the nodes due to anti-affinity rule. I just wondering if this could help to run multiple load test with current system in the prod cluster, i.e. we could provide instruction to create the namespace for each user, let each load test run in their own namespace, and then clean up the namespace afterwards. Just a thought....

codeblooded:
@wanlin31 It could provide some isolation from seeing other's tests, ensuring that they don't delete each other's resources, but it provides _no isolation_ in actually scheduling. It may hurt our metrics. For instance, your server and my client may both be scheduled on the same machine if they were in a separate namespace. This happens, in my understanding, b/c anti-affinities are not honored or enforced in separate namespaces.

wanlin31:
As I read your commit again in #49, understood! :)

wanlin31 (approved):

paulosjca (approved):

PR #51, opened 2020-09-09
Fix cluster role for listing pods

codeblooded:
The manager needs to be able to perform CRUD operations on load tests and to create, read and delete pods. The cluster role being generated was insufficient, because it scoped pods to v1. This change updates the role, allowing the manager to work with all pods, regardless of versions.

---

I do not know why it is querying beyond v1 for Pods. Their type is defined in the k8s.io/api/core/v1 package.  There may be a better fix for the future, such as limiting pods to a specific version when listing them. This is something to investigate and improve upon. For now, this fix suffices to deploy to a cluster with proper RBAC.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,6 +7,24 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods/status
+  verbs:
+  - get
+- apiGroups:
   - e2etest.grpc.io
   resources:
   - loadtests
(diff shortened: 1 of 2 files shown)


config/rbac/role.yaml
```diff
@@ -7,9 +7,9 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
-  - e2etest.grpc.io
+  - ""
```
codeblooded, line 10: Yes, it looks weird but `""` works for core resources, like pods.

config/rbac/role.yaml
```diff
@@ -19,17 +19,15 @@ rules:
   - update
   - watch
 - apiGroups:
-  - e2etest.grpc.io
+  - ""
   resources:
-  - loadtests/status
+  - pods/status
   verbs:
   - get
```
wanlin31, line 26: I totally agree that we should limit the privilege of each role to its minimum requirement, I am just wondering PodStatus is system generated information, we probably have no control over it except get, why did we set patch and update action for a status object?

wanlin31 (approved):

config/rbac/role.yaml
codeblooded: We didn't. The `pods/status` resources only declares the `get` verb.

codeblooded: We have `get`, `patch` and `update` verbs on the `loadtests/status` resource.

paulosjca (approved):

PR #52, opened 2020-09-09
Reorganize project, removing pkg and internal directories

codeblooded:
These commits rearrange the YAML files, also unifying config logic by moving the defaults into the config directory. It also removes the internal directory, since the entire project should be considered "internal".


config/rbac/bindings/component_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: component-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: pod-viewer-role
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: default
```
wanlin31, line 12: The pod-viewer is a ClusterRole, I am assuming this is for maintenance purpose, user given this role would be able to see all pods across the namespaces.

wanlin31 (approved):

config/rbac/bindings/component_role_binding.yaml
codeblooded: Yes, all our roles are at the cluster scope for the moment.

paulosjca (approved):

PR #53, opened 2020-09-11
Revise the load test status

codeblooded:
This change modifies the types and fields for a load test status. It sets minimal details in accordance with the Kubernetes API Conventions; meanwhile, providing enough information to debug issues and cleanup resources.


wanlin31 (approved):
Great! I think this make it clear about what status of the loadtest is!

paulosjca (approved):

PR #54, opened 2020-09-15
Move controller constants to config package

codeblooded:
This commit moves most constants that are declared in the controller into the config package. They are alphabetically sorted in a new constants.go file. This change allows the constants to be used across packages.

### DIFF
--- a/config/constants.go
+++ b/config/constants.go
@@ -17,26 +17,78 @@ limitations under the License.
 package config
 
 const (
-	// LoadTestLabel is a label which contains the test's unique name.
-	LoadTestLabel = "loadtest"
-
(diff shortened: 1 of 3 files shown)


codeblooded:
@grpc/test-infra This change makes it possible to begin using these constants outside of the controller package. This is useful for our new packages. Without this change, there would be duplication or cyclic dependencies.

wanlin31 (approved):

paulosjca (approved):

codeblooded:
As discussed with @paulosjca offline, this change allows us to remove some duplication. For example, the ready init container and controller can both use some of these constants.

PR #55, opened 2020-09-15
Add status package with PodsForLoadTest function

codeblooded:
This change introduces a status package meant to compute and convey the current state of the world. In addition, it adds a PodsForLoadTest helper function in a relationships.go file. This function uses the label relationships between pods and load tests to isolate pods that belong to a specific load test.

Merge #54 first.

### DIFF
--- a/config/constants.go
+++ b/config/constants.go
@@ -17,26 +17,78 @@ limitations under the License.
 package config
 
 const (
-	// LoadTestLabel is a label which contains the test's unique name.
-	LoadTestLabel = "loadtest"
-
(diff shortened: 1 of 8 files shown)


codeblooded:
@wanlin31 I know we talked about a separate package for CheckMissingPods and PodsForLoadTest. I had some second thoughts. Both seem related to the current status of the world. What is missing? What pods are running that belong to this load test? I felt like I should generalize the status package to be the "current state of the world" logic. What are your thoughts?

wanlin31 (approved):

paulosjca (approved):

PR #59, opened 2020-09-17
Update the status of a load test and prevent restarts

codeblooded:
In order to provide success and failure information to a user via a load test resource, the controller needs to diagnose the health at three levels. The container level provides insights about a particular process and its exit code, such as a failure to compile. The pod level provides insights about a pod using the aggregate health of its containers. The load test provides the status of the test using the aggregate health of its pods.

This pull request includes a series of commits to add observations at these three levels. In addition, it adds Reason strings to the load test API. Reason strings are designed to be a part of your API and to be consumed by machines. These reason strings explain the observations that led to the assigned health. Finally, it uses the previous observations to update the status of the load test and to prevent restarts after completion.

Examples of statuses produced by this change:

Screenshot 2020-09-18 at 11 22 58 AM
Screenshot 2020-09-18 at 11 25 40 AM
Screenshot 2020-09-18 at 11 31 35 AM

### DIFF
--- a/api/v1/loadtest_types.go
+++ b/api/v1/loadtest_types.go
@@ -243,6 +243,18 @@ const (
 	Errored = "Errored"
 )
 
+// InitContainerError is the reason string when an init container has failed on
+// one of the load test's pods.
+var InitContainerError = "InitContainerError"
+
+// ContainerError is the reason string when a container has failed on one of the
+// load test's pods.
+var ContainerError = "ContainerError"
+
+// PodsMissing is the reason string when the load test is missing pods and is still
+// in the Initializing state.
+var PodsMissing = "PodsMissing"
+
 // LoadTestStatus defines the observed state of LoadTest
 type LoadTestStatus struct {
 	// State identifies the current state of the load test. It is
--- a/api/v1/loadtest_types.go
+++ b/api/v1/loadtest_types.go
@@ -243,6 +243,12 @@ const (
 	Errored = "Errored"
 )
 
+// IsTerminated returns true if the test has finished due to a success, failure
+// or error. Otherwise, it returns false.
+func (lts LoadTestState) IsTerminated() bool {
+	return lts == Succeeded || lts == Failed || lts == Errored
+}
+
 // InitContainerError is the reason string when an init container has failed on
 // one of the load test's pods.
 var InitContainerError = "InitContainerError"
--- a/status/status.go
+++ b/status/status.go
@@ -16,8 +16,10 @@ limitations under the License.
 package status
 
 import (
+	"fmt"
 	"strings"
 
+	grpcv1 "github.com/grpc/test-infra/api/v1"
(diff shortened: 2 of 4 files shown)


gnossen (approved):
Not a detailed review, but LGTM to the overall idea. How would one see these statuses? `kubectl describe loadtest my-load-test`?

api/v1/loadtest_types.go
```diff
@@ -243,6 +243,24 @@ const (
 	Errored = "Errored"
 )
 
+// IsTerminated returns true if the test has finished due to a success, failure
+// or error. Otherwise, it returns false.
+func (lts LoadTestState) IsTerminated() bool {
+	return lts == Succeeded || lts == Failed || lts == Errored
+}
```
wanlin31, line 250: Like this function, it will be so useful for garbage collection :)

status/status.go
```diff
@@ -0,0 +1,170 @@
...
+
+		if contState == Errored {
+			message := fmt.Sprintf("container %q terminated with exit code %d", contStat.Name, *exitCode)
+			return Errored, grpcv1.ContainerError, message
+		}
+
+		if contState != Pending {
+			podState = contState
+		}
+	}
```
wanlin31, line 99: I think it is possible to have multiple run containers in pod.Containers, this part probably changes the podState every time it sees something different from pending, say we have runtime container array with status as success, success, pending, pending, we would have success as the over all podState, right?  For our case, I think it is absolute okay, just wondering.

wanlin31 (approved):

status/status.go
codeblooded: @wanlin31, very fair catch! As of right now, there is only 1 container so it works. However, we want anything other than Success to bubble up, not Pending. So, that way if we encounter Success, Pending, Pending... we get Pending, not success. Will push a quick fix

status/status.go
```diff
@@ -0,0 +1,67 @@
...
+	// succeeding or failing.
+	Pending State = "Pending"
+
+	// Succeeded indicates that the resource has terminated successfully,
+	// marked by a zero exit code.
+	Succeeded State = "Succeeded"
+
+	// Errored indicates that the resource has terminated unsuccessfully,
+	// marked by a non-zero exit code.
+	Errored State = "Failed"
```
paulosjca, line 43: Should be "Errored".

status/status_test.go
```diff
@@ -194,6 +194,23 @@ var _ = Describe("StateForPodStatus", func() {
 			Expect(state).To(Equal(Errored))
 			Expect(reason).To(Equal(grpcv1.ContainerError))
 		})
+
+		It("marks a pod as pending if all containers have not finished", func() {
```
paulosjca, line 198: I would reword as "if not all containers have finished."

paulosjca (approved):

PR #63, opened 2020-09-30
Improve Makefile and deployment process

codeblooded:
This change improves the Makefile by adding targets for each of the container images, renaming ambiguous targets, and adding environment variables to customize builds. It also removes an unnecessary dependency on the "test" target.

In addition, it improves the deployment process by adding introducing a way to dynamically generate a config file. In particular, it:

 - Introduces the config/cmd/configure.go program. This program uses a template file and environment variables to produce a config file.

 - Adds the config/defaults_template.yaml file as the default configuration template.

 - Removes the old config/example_config.yaml file, since it is no longer applicable.

 - Adds a "configure" phony target to the Makefile. This target invokes the configuration program with the default configuration template. It creates a config/defaults.yaml file, which is used by the controller.


wanlin31 (approved):

codeblooded:
Converting this back to draft PR for the time-being. Versioning the driver image on grpc/grpc tags alone is inadequate. Consider the scenario where we tag the driver at grpc release v1.32.0. Then, we make a change to the driver Dockerfile in grpc/test-infra. If the version of grpc remains unchanged, we would have the same tag. Converting this to draft allows me to amend and rethink the tagging strategy. Otherwise, I think this is a huge improvement.

codeblooded:
Reopening now that tags have been addressed.

codeblooded:
Closing and splitting as requested by @paulosjca.

PR #67, opened 2020-10-07
Add bazel cache volume

codeblooded:
The driver and bazel-built workers were failing, because the official bazel image has a complex output behavior. Bazel compiles into a bazel cache directory. Then, it creates links in the bazel-bin and bazel-out directories to the executables in the bazel cache.

This link structure does not allow the build init container to deliver an executable to the run container. The run container lacks the bazel cache available to the build init container. This prevents any executables from being accessible in the run container.

To resolve this issue, this commit introduces a bazel cache volume on all pods. It mounts the volume in the build init container and run container. When languages do not use bazel, the mounted directory is empty and has no effect. When languages use bazel, the mounted directory shares the bazel cache. This allows the bazel-bin and bazel-out links to resolve, making executables accessible to the run container.

### DIFF
--- a/config/constants.go
+++ b/config/constants.go
@@ -17,6 +17,15 @@ limitations under the License.
 package config
 
 const (
+	// BazelCacheVolumeName holds the name of the volume which allows images to
+	// share a bazel cache.
+	BazelCacheVolumeName = "bazel-cache"
+
+	// BazelCacheMountPath stores the directory where the bazel cache resides. For
+	// a description of the bazel image and its cache/output directories, see
+	// https://docs.bazel.build/versions/master/output_directories.html.
+	BazelCacheMountPath = "/root/.cache/bazel"
+
 	// BigQueryTableEnv specifies the name of the env variable that holds the name
 	// of the table where results should be written.
 	BigQueryTableEnv = "BQ_RESULT_TABLE"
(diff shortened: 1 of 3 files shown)


codeblooded:
@grpc/test-infra Will want to follow this PR up with a PR to increase the timeout of ready init container image. Both @wanlin31 and I noticed that the timeout of 5 minutes is too short for C++ and Python.

gnossen:
Bazel produces a hermetic artifact for *most* of its supported languages, with interpreted languages like Python and Ruby being an exception. I don't think you need to keep the entire Bazel worktree, as that will include many intermediary artifacts. For C++, Go, and Java, can you not just `cp $(realpath $ARTIFACT) destination`?

paulosjca (approved):

codeblooded:
@gnossen At the moment, we are compiling specific targets, like `//test/cpp/qps:qps_worker`. We do not own the build container, since we use the official bazel image. We don't want to create and maintain a new build container at the moment and wrap the gradle executable to do the copy. I believe we would need to do this, since the build occurs in an init container and is used in a non-init container. Correct me if I'm mistaken. After some discussion, we felt we can consider this as a potential optimization for the future.

codeblooded:
Merged so C++ and Python builds can be supported.

PR #68, opened 2020-10-08
Change ready init container timeout to 15 minutes

codeblooded:
The ready init container was set to timeout after 5 minutes, marking the test as erroring. Unfortunately, this is not enough time to build the C-based gRPC through bazel. On 2020-10-07, the C++ example scenario took 7 minutes and 8 seconds to compile off gRPC at master and run the test. I would expect that a slow build may exceed 10 minutes. For these reasons, this commit raises the timeout of the ready init container process to 15 minutes.

### DIFF
--- a/containers/init/ready/ready.go
+++ b/containers/init/ready/ready.go
@@ -40,7 +40,7 @@ const TimeoutEnv = "READY_TIMEOUT"
 
 // DefaultTimeout specifies the amount of time to wait for ready pods if the
 // environment variable specified by the TimeoutEnv constant is not set.
-const DefaultTimeout = 5 * time.Minute
+const DefaultTimeout = 15 * time.Minute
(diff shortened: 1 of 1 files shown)


wanlin31 (approved):

wanlin31:
I am wondering did you update the ready image in the container registry?

codeblooded:
> I am wondering did you update the ready image in the container registry?

@wanlin31 No, I have not done so. I don't want to push an image that isn't merged.

This brings up a separate discussion for a tagging strategy for images that are not merged. We could tag them with PR numbers, perhaps. /cc @paulosjca

paulosjca (approved):

codeblooded:
Amended commit message to improve it a little. No functional change. Merging.

PR #69, opened 2020-10-08
Add container image targets to Makefile

codeblooded:
This pull request adds make targets to build, tag and push container images, individually or in aggregate. It contains half of the changes of grpc/test-infra#63, split at the request of @paulosjca.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,15 @@
-
+# Version tag for all images
+TEST_INFRA_VERSION ?= "latest"
+# Version of the gRPC driver
+DRIVER_VERSION ?= "master"
+# Prefix for all images used as init containers, enabling use with registries
+# other than DockerHub
+INIT_IMAGE_PREFIX ?= ""
+# Prefix for all images used as runtime containers, enabling use with registries
+# other than DockerHub
+IMAGE_PREFIX ?= ""
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,15 @@
-
+# Version tag for all images
+TEST_INFRA_VERSION ?= "latest"
```
paulosjca, line 2: This may be acceptable, but I wonder if we should require the user to specify a version.

paulosjca (approved):

Makefile
paulosjca: I mean, specify a version and not default to "latest" (people will do it if you let them).

codeblooded: Good thing to think about. This just adopts behavior of kubebuilder's image where it defaults to latest. I'm afraid to make it an empty string. It is failing right now because of the colon suffix, like `docker push controller:`. Albeit, not always helpful errors. If it was `docker push controller`, then docker would automatically add the latest tag.

If the user sets no environment variables, it will tag the images as latest. However, it won't be able to push them to any registry. So, the cluster would be unable to run them. If I have no intention of pushing to GCR and am just quickly building the container on my machine, I may just use the latest image to run and then to discard it.

I agree with your previous feedback about requiring environment variables. I don't like that before running make either. If they plan to deploy, they have no choice... since the system will not deploy or work without them. If they do not plan to deploy (or they are deploying somewhere other than GKE and GCR), the defaults may suffice for them.

PR #70, opened 2020-10-09
Add example load test for Python

wanlin31:
This change adds an example yaml file for a loadtest in the Python
language and an example json scenario file for the python example
loadtest. This change has also update the example config file with
python language.

### DIFF
--- a/config/example_config.yaml
+++ b/config/example_config.yaml
@@ -29,4 +29,4 @@ languages:
 
   - language: python
     buildImage: l.gcr.io/google/bazel:latest
-    runImage: gcr.io/grpc-testing/e2etest/runtime/python3.7
+    runImage: docker.pkg.github.com/grpc/test-infra/python
--- a/config/example_config.yaml
+++ b/config/example_config.yaml
@@ -26,3 +26,7 @@ languages:
   - language: go
     buildImage: golang:1.14
     runImage: docker.pkg.github.com/grpc/test-infra/go
+
+  - language: python
+    buildImage: l.gcr.io/google/bazel:latest
+    runImage: gcr.io/grpc-testing/e2etest/runtime/python3.7
--- a/config/samples/python_example_scenario.json
+++ b/config/samples/python_example_scenario.json
@@ -1,6 +1,7 @@
 {
   "scenarios": [
     {
+      "name": "python_example_scenario",
       "benchmark_seconds": 30,
       "client_config": {
         "async_client_threads": 1,
@@ -34,7 +35,6 @@
         },
         "threads_per_cq": 0
       },
-      "name": "python_example_scenario",
       "num_clients": 1,
       "num_servers": 1,
       "server_config": {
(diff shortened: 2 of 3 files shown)


wanlin31:
I don't know if I should add the source script of the python example scenario in the commit or somewhere else, I can modify the commit message if it should be stated here.

config/samples/python_example_scenario.json
```diff
@@ -0,0 +1,65 @@
...
+          }
+        },
+        "rpc_type": "STREAMING",
+        "security_params": {
+          "server_host_override": "foo.test.google.fr",
+          "use_test_ca": true
+        },
+        "threads_per_cq": 0
+      },
+      "name": "python_example_scenario",
```
codeblooded, line 37: Why don't we move the name to the top, so it is quickly visible?

```suggestion
      "name": "python_example_scenario",
      "benchmark_seconds": 30,
      "client_config": {
        "async_client_threads": 1,
        "channel_args": [
          {
            "name": "grpc.optimization_target",
            "str_value": "latency"
          }
        ],
        "client_channels": 1,
        "client_processes": 0,
        "client_type": "SYNC_CLIENT",
        "histogram_params": {
          "max_possible": 60000000000.0,
          "resolution": 0.01
        },
        "load_params": {
          "closed_loop": {}
        },
        "outstanding_rpcs_per_channel": 1,
        "payload_config": {
          "bytebuf_params": {
            "req_size": 0,
            "resp_size": 0
          }
        },
        "rpc_type": "STREAMING",
        "security_params": {
          "server_host_override": "foo.test.google.fr",
          "use_test_ca": true
        },
        "threads_per_cq": 0
      },
```

config/example_config.yaml
```diff
@@ -26,3 +26,7 @@ languages:
   - language: go
     buildImage: golang:1.14
     runImage: docker.pkg.github.com/grpc/test-infra/go
+
+  - language: python
+    buildImage: l.gcr.io/google/bazel:latest
+    runImage: gcr.io/grpc-testing/e2etest/runtime/python3.7
```
codeblooded, line 32: Let's not reference a specific path and registry, since it is an example:

```suggestion
    runImage: docker.pkg.github.com/grpc/test-infra/python
```

codeblooded (changes requested):
Looks good with a couple thoughts. I would keep the source script in the colab notebook, since it is not official code. We will need to workout how we handle scenarios in the future.

config/samples/python_example_scenario.json
wanlin31: I kept it as it was generated just to make sure that if others interested in generate this scenario themselves, they will have exactly the same result as we have here, but I did get the idea that the original way makes it hard to people to see the scenario name, I agree the change. :)

config/example_config.yaml
wanlin31: Got it :). Thank you.

codeblooded (approved):

paulosjca (approved):

PR #72, opened 2020-10-13
Add defaults file configuration tool and validation

codeblooded:
This commit adds a tool that generates a defaults configuration file based on a template YAML file. This allows a generic configuration to apply to all environments. Meanwhile, the tool can customize it for options that vary. In addition, it adds validation of the defaults at generation and runtime.

```
Usage: go run config/cmd/configure.go <template-file> <output-file>

Configure is an executable that generates a defaults file for the manager. It
accepts a template file and replaces placeholders with data that may change
based on where the manager and container images will live and run.

This configure tool accepts two arguments. The first is <template-file>, which
is the input YAML file with placeholders for string interpolation. The second is
<output-file>, which is the path to write the output on disk.

The string interpolation is based on Go's text/template package. See
https://pkg.go.dev/text/template for a description of the syntax. All flags
passed to the script (except -validate) are accessible within the template.

Flags:
  -driver-pool string
        pool where drivers are scheduled by default (required)

        Drivers will be scheduled by default on nodes with a "pool" label that
        matches this -driver-pool flag.
  -image-prefix string
        prefix to append to container images (optional)

        This -image-prefix flag allows a specific prefix to apply to all
        container images that are not used as init containers.
  -init-image-prefix string
        prefix to append to init container images (optional)

        This -init-image-prefix flag allows a specific prefix to apply to all
        init container images.
  -validate
        validate the output configuration for correctness (default true)
  -version string
        version of all docker images to use (default "latest")
  -worker-pool string
        pool where workers are scheduled by default (required)

        Workers will be scheduled by default on nodes with a "pool" label that
        matches this -worker-pool flag.
```


paulosjca (commented):
It is not clear how this configuration tool would be used. Would it generate a configuration file that would then be used to deploy the controller to the cluster? Perhaps there is a missing make target that goes with it.

paulosjca:
From offline conversation, the intention is that the tool would be used independently of the makefile.

paulosjca (approved):

PR #73, opened 2020-10-14
Add flag for reconcile timeouts

codeblooded:
This change removes a magic value and unused constant that were both
intended to set a timeout for a single invocation of the Reconcile
method. It replaces them by adding a `-reconcile-timeout` flag on the
manager. This flag's value is provided to the LoadTestReconciler upon
initialization.

In addition, this commit sets the flag to 2 minutes for the deployment
of the manager and its controller. This number is consistent with the
previous magic value.

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -23,3 +23,4 @@ spec:
         args:
         - "--metrics-addr=127.0.0.1:8080"
         - "--enable-leader-election"
+        - "--reconciliation-timeout=2m"
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -34,16 +34,13 @@ import (
 	"github.com/grpc/test-infra/config"
 )
 
-// reconcileTimeout specifies the maximum amount of time any set of API
-// requests should take for a single invocation of the Reconcile method.
-const reconcileTimeout = 1 * time.Minute
-
 // LoadTestReconciler reconciles a LoadTest object
 type LoadTestReconciler struct {
 	client.Client
 	Defaults *config.Defaults
 	Log      logr.Logger
 	Scheme   *runtime.Scheme
+	Timeout  time.Duration
 }
 
 // +kubebuilder:rbac:groups=e2etest.grpc.io,resources=loadtests,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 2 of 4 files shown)


controllers/loadtest_controller.go
```diff
@@ -55,12 +52,19 @@ type LoadTestReconciler struct {
...
+	var cancel context.CancelFunc
 	var err error
 
-	log := r.Log.WithValues("loadtest", req.NamespacedName)
-	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	if r.Timeout == nil {
+		ctx, cancel = context.WithCancel(context.Background())
+	} else {
+		ctx, cancel = context.WithTimeout(context.Background(), *r.Timeout)
+	}
```
wanlin31, line 63: Just want to make sure I understand this right: if use did not provide a timeout, we will not reconcile?

codeblooded: @wanlin31 no. If users do not specify a timeout flag, we create a context without a timeout. All contexts have a cancellation function. If invoked, the context expires. We defer this cancellation, meaning it occurs after the Reconcile completes. Contexts are hard to explain, but the docs on context are helpful: https://pkg.go.dev/context#Context.

Effectively, there is no timeout if the flag is not set. We will probably change the Deployment of the controller to include it. Why not impose a timeout? Kubebuilder defaults to no timeout in generated code, and timeouts make it virtually impossible to debug complex issues in a debugger. For development, they are a hindrance. For prod, they have a purpose.

wanlin31: Got it! Thank you!

wanlin31 (approved):

controllers/loadtest_controller.go
wanlin31: What would be the possible reason for a timeout? Is it safe to run reconcile without a time out? Will it trying forever to reconcile something that is not possible?

codeblooded: @wanlin31 > What would be the possible reason for a timeout?

When we make calls to an external API, in this case Kubernetes, we pass the timeout. Essentially, if the response does not come back before the context is done (timeout expired or cancelled), the client hangs up. It says, "If kubernetes takes too long to respond, forget it. I don't want to be blocking." Contexts should be set for any external request we make with Go on a production process. 

> Is it safe to run reconcile without a time out?

Yes, this is particularly true in two cases: development and testing. If the Kubernetes API hangs, we just wait. This is desirable for testing, because ginkgo should have a built in timeout mechanism. For development it’s even more desirable because of how contexts work.

When setting time limits, some languages use timeouts and some use deadlines. Go contexts appear deadline-driven. Consider we create a timeout context for 2 minutes. I use it to call an API (A). I then use it to call another API (B). Say (A) takes 1.5 minutes to respond and (B) takes 0.75 minutes to respond. In this situation, (B) would always fail. It’s not a timeout that is reset, it is a deadline. It would only work if (A)+(B) < ctx.Deadline(). This makes debugging a nightmare. Your deadline is reached before you hit the breakpoint often. 

> Will it trying forever to reconcile something that is not possible?

No, this is just waiting for a connection and response from an API. If it is impossible, the API should error. This could hang forever though when I unset.

codeblooded:
@paulosjca Renamed flag `-timeout` to `-reconcile-timeout` and amended the configuration YAML so that the previous magic value, 2 minutes, applies to future deployments.

codeblooded:
Fixed commit message which referenced old flag name.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -23,3 +23,4 @@ spec:
         args:
         - "--metrics-addr=127.0.0.1:8080"
         - "--enable-leader-election"
+        - "--reconcile-timeout=2m"
```
paulosjca, line 26: I think reconciliation-timeout would be clearer. reconcile-timeout gives the impression that timeouts are what is being reconciled.

paulosjca (approved):

PR #75, opened 2020-10-16
Format example test configurations and comment out invalid references to BigQuery

paulosjca:
This change formats the example test configurations with "prettier" and comments out references that point to invalid locations in BigQuery. Commenting out invalid references allows the examples to be used out of the box.


config/samples/java_example_loadtest.yaml
```diff
@@ -1,64 +1,75 @@
...
+        repo: https://github.com/grpc/grpc-java.git
+        gitRef: master
+      build:
+        command: ["gradle"]
+        args:
+          [
+            "-PskipAndroid=true",
+            "-PskipCodegen=true",
+            ":grpc-benchmarks:installDist",
+          ]
```
codeblooded, line 35: This is a very interesting choice for prettier. Why did it keep the brackets? It could have been:

```
args:
  - "-PskipAndroid=true"
  - "-PskipCodegen=true"
  - ":grpc-benchmarks:installDist"
```

codeblooded (approved):
LGTM

config/samples/java_example_loadtest.yaml
paulosjca: The design decision to use brackets or not was made by the user. Prettier reformatted the file to break up a long line, but kept the original decision. This behavior makes it easier for the user to override the change if they don't agree with it, by moving the items together into one line with only whitespace changes. I think this is a good choice of default behavior by the tool.

PR #76, opened 2020-10-16
Send quit RPC to terminate workers after test

wanlin31:
Previously, after test finishes, the driver will have a status of
complete but the workers will keep running and prevent the next
pod to be assigned to the nodes they are running on. This
change let the driver send a quit RPC to terminate the workers
after the test, once the workers are terminated they would have a
status of complete as the driver. The nodes only have complete
pods are available for next pod assignment.


codeblooded (approved):
LGTM 🙂

paulosjca (approved):

PR #79, opened 2020-10-22
Optimize flow of execution for test reconciliation

codeblooded:
The previous flow of execution for the (*LoadTestReconciler).Reconcile method included pre-emptively fetching resources and making unnecessary calls to account for defaults. Aside from the unnecessary requests, the flow was hard to follow.

This change reorders the method, removing unnecessary calls and switching to fetch resources lazily.

### DIFF
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -70,6 +70,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	rawTest := new(grpcv1.LoadTest)
 	if err = r.Get(ctx, req.NamespacedName, rawTest); err != nil {
 		log.Error(err, "failed to get test", "name", req.NamespacedName)
+		// do not requeue, the test may have been deleted or the cache is invalid
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
@@ -57,6 +58,8 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	var cancel context.CancelFunc
 	var err error
 
+	log := r.Log.WithValues("loadtest", req.NamespacedName)
+
 	if r.Timeout == 0 {
 		ctx, cancel = context.WithCancel(context.Background())
 	} else {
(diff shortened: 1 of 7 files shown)


controllers/loadtest_controller.go
```diff
@@ -56,64 +57,44 @@ type LoadTestReconciler struct {
...
-		log.Error(err, "failed to list pods", "namespace", req.Namespace)
-		return ctrl.Result{Requeue: true}, err
-	}
-
-	loadtests := new(grpcv1.LoadTestList)
-	if err = r.List(ctx, loadtests); err != nil {
-		log.Error(err, "failed to list loadtests")
-		return ctrl.Result{Requeue: true}, err
-	}
-
```
wanlin31, line 86: I like that you removed this part, I used to think all the time why we have fetch all the LoadTest and  fetch the one that actually triggered the reconciliation later. Make more scene and more efficient this way.

controllers/loadtest_controller.go
```diff
@@ -56,64 +57,44 @@ type LoadTestReconciler struct {
...
 	loadtest := new(grpcv1.LoadTest)
 	if err = r.Get(ctx, req.NamespacedName, loadtest); err != nil {
 		log.Error(err, "failed to get loadtest", "name", req.NamespacedName)
-
-		// do not requeue, the load test may have been deleted
```
wanlin31, line 91: why remove the comments? I found them pretty useful :)

controllers/loadtest_controller.go
```diff
@@ -56,64 +57,44 @@ type LoadTestReconciler struct {
...
+	if err = r.Defaults.SetLoadTestDefaults(loadtestWithDefaults); err != nil {
+		log.Error(err, "failed to clone test with defaults")
 		return ctrl.Result{}, err
 	}
-	if err = r.Update(ctx, loadtest); err != nil {
-		log.Error(err, "failed to update loadtest with defaults")
+	if !reflect.DeepEqual(loadtest, loadtestWithDefaults) {
+		if err = r.Update(ctx, loadtestWithDefaults); err != nil {
+			log.Error(err, "failed to update test with defaults")
+			return ctrl.Result{}, err
```
wanlin31, line 84: Little confused about the update function, I used to think it happens like this: every time we get a triggering LoadTest -> we made sure that it follows the default -> we update the triggering LoadTest. I understand you make the change to make it a lazy update, but say if we do found a difference (that is line 82), why are we updating the `loadtestWithDefaults` instead of `loadtest`?

codeblooded: `loadtestWithDefaults` is equivalent to `loadtest`, except it adds defaults. It's name, namespace, and GVK (group, version, kind) information is identical. If it looks like a duck... walks like a duck... quacks like a duck... it is a duck. To Kubernetes, this is enough for it to know which object we are talking about (even if it  is not the same object in memory). It will be treated the same. We don't want to update with loadtest, because that would do nothing. It has no defaults. They both are kept around, b/c line 81 ensures that we do not update every single time we reconcile (only when we need to set defaults).

controllers/loadtest_controller.go
codeblooded: Good to know. I did not realize they were useful. I thought they may be redundant since we had a comment every time we reconciled. I can see it can be useful in some contexts.

codeblooded:
Moving to draft, because I will rename some of these variables and improve it a little. Will move back to ready when fixed and re-request review. I think this will aid with clarity.

controllers/loadtest_controller.go
wanlin31: " To Kubernetes, this is enough for it to know which object we are talking about (even if it is not the same object in memory)" Got it ! Thx. :)

wanlin31 (approved):

codeblooded:
@wanlin31 I added back the comment that you liked. I think this should be good to go. @paulosjca do you want to review?

paulosjca (approved):

PR #82, opened 2020-10-30
Optimize branch checkouts in clone init container

codeblooded:
This change optimizes the checkout of a branch or tag for the clone init container. Previously, repositories were cloned off the master branch. This commit switches to cloning a specific, single-branch only.

Unfortunately, this breaks checking out a commit hash, because a tree must be downloaded for git to resolve a specific commit.

### DIFF
--- a/containers/init/clone/clone.sh
+++ b/containers/init/clone/clone.sh
@@ -17,5 +17,4 @@ set -ex
 
 cd /src/workspace
 ls -A | xargs -r rm -fr
-git clone --recursive $CLONE_REPO .
-git checkout $CLONE_GIT_REF
+git clone --recursive $CLONE_REPO --branch $CLONE_GIT_REF --single-branch .


codeblooded:
This addresses some feedback received, but it introduces a new problem. Not being able to clone/checkout a specific commit seems problematic to me. This is especially true for CI builds. When batching them together -- like every merged change in the last 2 hours) -- you can only compare using commit hashes.

In my opinion, this change AS IS should be marked as a wontfix. If we decide to pursue this optimization, we should no longer use the gitRef option. It should be broken up into `branch`, `tag` or `commit`. Then, we should introduce additional logic to the init container to select the `commit`. This may be a better design anyway. For all repos, aside from grpc/grpc, the speed up will not be super significant. It just becomes an annoyance.

codeblooded:
Closing in favor of #86.

PR #83, opened 2020-10-30
Remove Failed status from load tests

codeblooded:
This change removes the Failed state from a load test. It replaces it with the Errored state.


codeblooded:
An enhancement should be create to improve error messages. This will occur in a separate PR.

paulosjca (approved):

PR #86, opened 2020-11-05
Remove unnecessary checkouts when cloning

codeblooded:
The clone docker image is used as an init container which fetches the code at a specific commit, tag or branch. Previously, it used the git-clone command, which clones the master branch by default. Developers wanted to avoid the unnecessary checkout when they are testing off another branch or tag.

The git-clone command does provide the --branch and --single-branch flags, but these options are incompatible with commits due to a server-side setting that GitHub has not enabled. See https://stackoverflow.com/questions/3489173/how-to-clone-git-repository-with-specific-revision-changeset.

This commit achieves similar behavior, but it uses an empty git repository, fetches a remote and resets to a specific branch, tag or commit. This avoids the unnecessary checkout.

### DIFF
--- a/containers/init/clone/clone.sh
+++ b/containers/init/clone/clone.sh
@@ -14,8 +14,11 @@
 # limitations under the License.
 
 set -ex
-
 cd /src/workspace
 ls -A | xargs -r rm -fr
-git clone --recursive $CLONE_REPO .
-git checkout $CLONE_GIT_REF
+git init
+git remote add origin $CLONE_REPO
+git fetch
+git reset --hard $CLONE_GIT_REF
+git submodule update --init --recursive
+chmod -R 777 .


paulosjca (approved):

containers/init/clone/clone.sh
```diff
@@ -14,8 +14,11 @@
...
 cd /src/workspace
 ls -A | xargs -r rm -fr
-git clone --recursive $CLONE_REPO .
-git checkout $CLONE_GIT_REF
+git init
+git remote add origin $CLONE_REPO
+git fetch
+git reset --hard $CLONE_GIT_REF
+git submodule update --init --recursive
+chmod -R 777 .
```
jtattermusch, line 24: why is the chmod required?

containers/init/clone/clone.sh
```diff
@@ -14,8 +14,11 @@
 # limitations under the License.
 
 set -ex
-
 cd /src/workspace
 ls -A | xargs -r rm -fr
-git clone --recursive $CLONE_REPO .
-git checkout $CLONE_GIT_REF
+git init
```
jtattermusch, line 19: nit: leave a comment that this seemingly weird way of cloning the repo is done for a good reason (and describe the problem it's avoiding).

containers/init/clone/clone.sh
```diff
@@ -14,8 +14,11 @@
...
+git remote add origin $CLONE_REPO
+git fetch
```
jtattermusch, line 21: nit: `git fetch origin`?

containers/init/clone/clone.sh
codeblooded: I don't make assumptions that the container will only be used with kubernetes. In fact, I test containers using docker alone before I ever throw them at kubernetes. That was the case here.

When mounting a pure docker volume and running this container, the resulting repository has inaccessible permissions. The chmod ensures that the contents written to the volume are mutable on both ends of the volume.

In kubernetes, immutability with volumes will be reintroduced soon as a property of the Volume type. It is currently an alpha feature. By default, kubernetes allows volumes to be read/write. This masks the problem under the hood.

jtattermusch: The concern is that git repos actually persist some mode information along with the file (e.g. the executable flag for scripts). So it feels wrong that you check out a repo and immediately overwrite all the information provided by git.
(e.g. non-executable scripts are changed into executable scripts which can lead to a different behavior when building locally vs when building in kubernetes - imagine someone forgets to set an executable flag for their script but things magically work in the oss benchmarks). At the very least you should provide a comment why the `chmod -R 777 .` is useful here.

PR #90, opened 2020-11-12
Add comments and address clone container feedback

codeblooded:
This change adds comments to the containers/init/clone/clone.sh script, explaining the process. In addition, it makes some minor changes that were requested by @jtattermusch in #86.

### DIFF
--- a/containers/init/clone/clone.sh
+++ b/containers/init/clone/clone.sh
@@ -16,9 +16,21 @@
 set -ex
 cd /src/workspace
 ls -A | xargs -r rm -fr
+
+# This process initializes an empty git repository, adds and fetches objects
+# from the $CLONE_REPO, checks out the $CLONE_GIT_REF and then updates the
+# submodules. This prevents the unnecessary checkout of the master branch.
+# This process is similar to other CI systems, including GitHub actions. See:
+# https://stackoverflow.com/questions/3489173.
+
 git init
 git remote add origin $CLONE_REPO
-git fetch
+git fetch origin
 git checkout $CLONE_GIT_REF
 git submodule update --init --recursive
+
+# At this point, the files and the directory are read-only when used with a
+# Docker volume. The mode is changed to ensure that consumers of the directory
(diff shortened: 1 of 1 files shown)


jtattermusch (approved):
LGTM, thanks!

containers/init/clone/clone.sh
```diff
@@ -16,9 +16,23 @@
 set -ex
 cd /src/workspace
 ls -A | xargs -r rm -fr
+
+# The git clone command clones code off the master branch by default or a branch
```
paulosjca, line 20: I would rewrite it to focus on the problem that is being solved:

The git clone command clones code off the master branch by default or a branch or tag specified through the flag --branch. However, the flag --branch is not supported for commits to GitHub repositories, so the default would be to (unnecessarily) clone the master branch and then checkout the desired branch.

The process below initializes an empty git repository, adds and fetches objects from $CLONE_REPO, checks out $CLONE_GIT_REF and then updates the submodules. This process works around the need to clone an unnecessary branch.

containers/init/clone/clone.sh
```diff
@@ -16,9 +16,27 @@
...
```
paulosjca, line 20: I think the explanation becomes clearer if it is limited to what is actually happening, and points to where the idea came from. Something like this (deleting the first paragraph completely – additional discussion might go to the description of the commit of PR, if useful):

```shell
# This process initializes an empty git repository, adds and fetches objects
# from the $CLONE_REPO, checks out the $CLONE_GIT_REF and then updates the
# submodules. This prevents the unnecessary checkout of the master branch.
# This process is similar to other CI systems, including GitHub actions. See:
# https://stackoverflow.com/questions/3489173.
```
I am not even sure about keeping the mention of "other CI systems", since there is no reference pointing to that, but it's ok to keep.

paulosjca: Ping (1 month).

paulosjca (approved):

PR #91, opened 2020-11-13
Add Timeout and TTL to LoadTest

wanlin31:
This commit adds Timeout and TTL to the LoadTest's schema and the
logics for controller to mark load test runs too long as error and
delete the LoadTest lives longer than TTL.


codeblooded:
This is likely to be a merge conflict with #92, so we'll need to coordinate what goes in first.

codeblooded (changes requested):
@wanlin31 There's a few thoughts I had so far. We can schedule a meeting to discuss some of them if you'd like. Overall, it's looking good 👍🏻

config/samples/go_example_loadtest.yaml
```diff
@@ -50,6 +50,12 @@ spec:
   # results:
   #   bigQueryTable: "example-project.foo.demo_dataset"
 
+  # timeout and ttl field only take numbers and converted to duration, the unit
+  # of duration is fixed as ns, for example: 1s = 1000000000 ns, the number
+  # below indicate timeout=30s, ttl=2min
+  timeout: 30000000000
+  ttl: 120000000000
```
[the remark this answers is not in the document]
wanlin31: I was going to talk about this in our meeting :). I made TTL and Timeout as time.duration, unfortunately go 1.14 fixes the unit as ns, see (https://github.com/golang/go/issues/4712), and it doesn't take scientific notion when I played with it. 

Method I : Direct fill as duration 
Pro: fail fast 
Con: inconvenient to use, prone to human error, but it might be okay since we give templates

Method II : Direct fill as string, and parse to durations
Pro: you can have you own units, convenient to use
Con: format check depend on controller

Method III : Direct fill as int64, and convert to durations
Pro: more convenient than Method I to some extend
        fail fast

When I trying to explain the good and bad of three ways, I tend to agree that take the field as int64 could be the best for the scenario. I will change it. :)

controllers/loadtest_controller.go
```diff
@@ -105,10 +143,14 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 	ownedPods := status.PodsForLoadTest(test, pods.Items)
-	test.Status = status.ForLoadTest(test, ownedPods)
+	test.Status, isNewlyScheduled, isNewlyTerminated = status.ForLoadTest(test, ownedPods)
+
 	if err = r.Status().Update(ctx, test); err != nil {
 		log.Error(err, "failed to update test status")
-		return ctrl.Result{Requeue: true}, err
+		if err = r.Delete(ctx, rawTest); err != nil {
+			log.Error(err, "fail to delete test", "name", req.NamespacedName)
+		}
```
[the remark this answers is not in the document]
wanlin31: My thought is like this, open to better/other ways to do that definitely: we calculate the start time and the end time within status function, failing to update these field has impact on when do we re-queue the LoadTest for Timeout and TTL. 

Method I: We either delete the LoadTest and let the user know there is a problem and ask users to do it again.
Pro: We would not have orphaned Loadtest
con: we might delete the LoadTest necessarily.

Method II: ignore it 
Pro: We will not delete the LoadTest when the problem is only related to kubernetes.
Con:this LoadTest could in fact timeout but not mark as error, or live forever. 

I put it here just as one way, I want to learn about how Paulo and you think.

controllers/loadtest_controller.go
```diff
@@ -73,7 +87,32 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	curTTL := rawTest.Spec.TTL
```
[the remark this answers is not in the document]
wanlin31: I will update this.

status/status.go
```diff
@@ -105,15 +106,30 @@ func StateForPodStatus(status *corev1.PodStatus) (state State, reason string, me
 // pods it owns. This sets the state, reason and message for the load test. In
 // addition, it attempts to set the start and stop times based on what has been
 // previously encountered.
-func ForLoadTest(test *grpcv1.LoadTest, pods []*corev1.Pod) grpcv1.LoadTestStatus {
+func ForLoadTest(test *grpcv1.LoadTest, pods []*corev1.Pod) (grpcv1.LoadTestStatus, bool, bool) {
```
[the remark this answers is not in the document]
wanlin31: I just need to have two flags to re-queue the LoadTest, best practice of doing this? 

Currently, timeout could be due to the test itself or the scheduling problem, adding TimeoutExceeded state tells you it is timed out, but you still don't know whether it is due to scheduling issue or the test itself, my feeling is this TimeoutExceeded state might not give a lot further information than error itself to users. I am not saying we should not add it, just some thoughts from a usage point of view.

controllers/loadtest_controller.go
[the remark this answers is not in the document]
wanlin31: I think about this again: 
If we delete the LoadTest failed to update at this moment, we could have delete the LoadTest unnecessarily only to save resource, it doesn't have to be this strict with saving resource. 

If we  just re queue it, the worst case scenario is that this LoadTest could live forever and its pods could be forever running and occupying the resource, if this happened a lot and the resource will be depleted, user interference will be necessary to clean up things like this, do not have to worry too much about this.

I convert it back to where it was like.

api/v1/loadtest_types.go
```diff
@@ -208,12 +208,10 @@ type LoadTestSpec struct {
...
 
 	// Timeout provides the longest running time allowed for a LoadTest.
-	// +optional
-	TimeoutSeconds *int32 `json:"timeoutSeconds,omitempty"`
+	TimeoutSeconds *int32 `json:"timeoutSeconds"`
 
 	//TTL provide the longest time a LoadTest could live on the luster.
-	// +optional
-	TTLSeconds *int32 `json:"ttlSeconds,omitempty"`
+	TTLSeconds *int32 `json:"ttlSeconds"`
```
codeblooded: I'm not sure these should be mandatory. I think our system should have sane defaults. These should not be pointers if they are required. The pointers convey optionality ([API Conventions: Optional vs. Required](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required)).

wanlin31: I will fix the pointer type if we still want it mandatory, thank for you pointing out!

I think whether we want to have them mandatory or not is really depend on what we do if these values are not set. My understanding is we definitely need these values to perform garbage collection.  
1. If we are going to give them default values if not set, we could definitely make them optional, but we would need to implement them in cluster configuration and check if they are actually there. Moreover I think these two values are more of properties of load tests than the cluster. 
2. If we are going to reject the load test without these values in controller, we might just reject it before we get to scheduling.

wanlin31:
I am still working on this Pr. Things I am working on:
1.  Just realized, even we made ttl and timeout mandatory, we still need to have checks to make sure they are reasonable values, and decide what to do if they are not.
2. I am moving the logics of calculating time for requeueAfter out, so that I could add test for it.
3. Thinking about if I could add better test for status.

api/v1/loadtest_types.go
```diff
@@ -206,6 +206,12 @@ type LoadTestSpec struct {
 	// Scenarios provides a list of configurations for testing.
 	// +optional
 	Scenarios []Scenario `json:"scenarios,omitempty"`
+
+	// Timeout provides the longest running time allowed for a LoadTest.
+	TimeoutSeconds int32 `json:"timeoutSeconds"`
+
+	//TTL provide the longest time a LoadTest could live on the luster.
```
codeblooded, line 213: I think there may be typos here, but otherwise good. 

```suggestion
	// TTL provide the longest time a LoadTest could live on the cluster.
```

config/samples/go_example_loadtest.yaml
```diff
@@ -50,6 +50,12 @@ spec:
   # results:
   #   bigQueryTable: "example-project.foo.demo_dataset"
 
+  # timeout and ttl field only take numbers and converted to duration, the unit
+  # of duration is fixed as ns, for example: 1s = 1000000000 ns, the number
+  # below indicate timeout=30s, ttl=2min
```
codeblooded, line 55: I like that we add these to the examples, but I wish we could elaborate a little more on each one.

Maybe we could break them into each having their own comments and explain what a TTL or timeout does to effect a test.

As a developer, I feel like I may be confused to know what to put for both of them. It may be helpful to specify the timeout is the maximum amount of time a test is allowed to run. Meanwhile, the TTL is the amount of time the load test and its result status should live in memory before being garbage collected.

If I was using this framework for the first time, I probably would not be likely to check the docs for these fields. I'd probably just play around with these examples and read their comments.

controllers/loadtest_controller.go
```diff
@@ -73,7 +72,29 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	loadTestTTL := time.Duration(rawTest.Spec.TTLSeconds) * time.Second
+	loadTestTimeout := time.Duration(rawTest.Spec.TimeoutSeconds) * time.Second
```
codeblooded, line 76: Great conversion.

nit: can we call this ttl and timeout or testTTL and testTimeout? We dropped the "load" from the test in a previous cleanup, because this is the LoadTestReconciler. The type of test is implied.

controllers/loadtest_controller.go
```diff
@@ -73,7 +72,29 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	if loadTestTTL == 0 {
+		log.Info("TTL is invalid", "TTL", loadTestTTL)
+	}
+
+	if loadTestTimeout == 0 {
+		log.Info("Timeout is invalid", "Timeout", loadTestTimeout)
+	}
+
+	if loadTestTimeout > loadTestTTL {
+		log.Info("loadTestTTL is less than loadTestTimeout", "Timeout", loadTestTimeout, "TTL", loadTestTTL)
```
codeblooded, line 87: nit: logs can be rendered as plain text or json. Let's make sure the fields, like the TTL all have the same case and name in all the log statements. I would suggest `ttl` or `testTTL`. JSON keys don't often start with caps.

controllers/loadtest_controller.go
```diff
@@ -73,7 +72,29 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	}
+
 	if rawTest.Status.State.IsTerminated() {
+		if time.Now().Sub(rawTest.Status.StartTime.Time) >= loadTestTTL {
+			log.Info("test expired, deleting", "startTime", rawTest.Status.StartTime, "loadTestTTL", loadTestTTL)
+			if err = r.Delete(ctx, rawTest); err != nil {
+				log.Error(err, "fail to delete test", "name", req.NamespacedName)
```
codeblooded, line 94: We can leave the name of the load test off, because it is automatically added to every log statement in reconcile. 

```suggestion
				log.Error(err, "fail to delete test")
```

controllers/loadtest_controller.go
```diff
@@ -139,12 +163,36 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		}
 	}
 
+	reQueueTime := getRequeueTime(previousStatus, test, log)
```
codeblooded, line 166: nit: let's keep the same case as the function. 

```suggestion
	requeueTime := getRequeueTime(previousStatus, test, log)
```

controllers/loadtest_controller_test.go
```diff
@@ -759,3 +762,87 @@ var _ = Describe("Pod Creation", func() {
 		})
 	})
 })
+
+var _ = Describe("getRequeueTime", func() {
+	var test *grpcv1.LoadTest
+	var reconciler LoadTestReconciler
```
codeblooded, line 768: A reconciler is a large object, can we make this a pointer?

```suggestion
	var reconciler *LoadTestReconciler
```

status/status.go
```diff
@@ -114,6 +115,16 @@ func ForLoadTest(test *grpcv1.LoadTest, pods []*corev1.Pod) grpcv1.LoadTestStatu
 		status.StartTime = test.Status.StartTime
 	}
 
+	timeout := time.Duration(test.Spec.TimeoutSeconds) * time.Second
+
+	// Here marked the LoadTest running too long as errored. This status update
+	// could trigger cleanup_agent to terminate its workers.
+	if time.Now().Sub(status.StartTime.Time) >= timeout {
+		status.StopTime = optional.CurrentTimePtr()
+		status.State = grpcv1.Errored
```
codeblooded, line 124: Great! We should add a reason string and message explaining the timeout. You can see how I did this in #92, adding KubernetesError reason and a string message. 

The reason could be something like `TimeoutExceeded`.

codeblooded (changes requested):
Looking very good. Very well done in my opinion. A few changes I'd like. 

I'm remarkably impressed by how much ginkgo and BDD tests come so naturally for you now.

config/samples/go_example_loadtest.yaml
codeblooded: We should also add them to other languages.

controllers/loadtest_controller.go
wanlin31: I wanted to use ttl and timeout originally, but I found that in the same func we have a Timeout for reconciler (https://github.com/grpc/test-infra/blob/5c7cfabd47b59d9f7c35494be247b31414a2d431/controllers/loadtest_controller.go#L45), I will change them to testTTL and testTimeout, that is no problem :) Thank you for coming up names.

controllers/loadtest_controller.go
wanlin31: I will check!

config/samples/go_example_loadtest.yaml
wanlin31: Yes, this just is an example for us to discuss if we want to do it the way it is now, once we are sure about this, I will add them to all other languages (go is the easiest/fastest to try). Thanks for reminding me!

For the explanation, I will separate them and explain more for each.

wanlin31:
I have updated the items we talked about yesterday. I think it is okay to be reviewed.

config/samples/cxx_example_loadtest.yaml
```diff
@@ -50,6 +50,21 @@ spec:
   # results:
   #   bigQueryTable: "example-project.foo.demo_dataset"
 
+  # timeoutSeconds field is to advise the longest time a test
+  # could possibly run, test runs longer than given value will be marked as
+  # Error and would not be allocate more resource. The field takes integer
```
codeblooded, line 55: nit: I think the state is past tense, Errored. I think this explains the details. I think @paulosjca may have feedback on wording.

controllers/loadtest_controller.go
```diff
@@ -139,12 +155,38 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		}
 	}
 
+	reQueueTime := getRequeueTime(test, previousStatus, log)
```
codeblooded, line 158: nit: I would keep consistent case with function and make the variable `requeueTime`.

controllers/loadtest_controller.go
```diff
@@ -139,12 +155,38 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 	return ctrl.Result{}, nil
 }
 
-// LoadTestMissing categorize missing components based on their roles at specific
-// moment. The struct is a wrapper to help us get role information associate
-// with components.
+// getRequeueTime takes a LoadTest and its previous state, and compares the
+// previous status of the load test with its updated status and return a
+// calculated requeue time. The reason we take the test itself other than
+// its current status is because the ttl and time out are saved in its spec.
```
codeblooded, line 168:
```suggestion
// its current status is because the ttl and timeout are saved in its spec.
```

controllers/loadtest_controller.go
```diff
@@ -139,12 +155,38 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+// if the load test has just been assigned a start time, getRequeueTime returns
+// the timeout specified within its spec. If the load test has been just
+// assigned stop time, getRequeueTime returns its ttl specified within its spec
+// less its actual running time. In other cases, returns a 0 value time.duration
```
codeblooded, line 172: nit: I think @paulosjca has reviewed me on spelling out small numbers before. He may want this to be changed from 0 to zero.

controllers/loadtest_controller.go
```diff
@@ -139,12 +155,38 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+func getRequeueTime(updatedLoadTest *grpcv1.LoadTest, previousStatus grpcv1.LoadTestStatus, log logr.Logger) time.Duration {
+	reQueueTime := time.Duration(0)
+
+	if previousStatus.StartTime == nil && updatedLoadTest.Status.StartTime != nil {
```
codeblooded, line 176: What would happen in the following case? The controller sets the start time and requeues, but then the controller crashes and restarts. The queue is lost. How would this logic be triggered again? I have no idea how we'd handle that case.

controllers/loadtest_controller_test.go
```diff
@@ -759,3 +762,87 @@ var _ = Describe("Pod Creation", func() {
...
+			expectedRequeueTime := time.Duration(0)
+			previousStatus := grpcv1.LoadTestStatus{StartTime: &startTime}
+			test.Status = grpcv1.LoadTestStatus{StartTime: &startTime}
+
+			requeueTime := getRequeueTime(test, previousStatus, reconciler.Log)
+
+			Expect(requeueTime).To(Equal(expectedRequeueTime))
+		})
+
+		It("returns 0 duration", func() {
```
codeblooded, line 824: It blocks should have different names. This allows us to see what failed quickly. So perhaps these should be more specific, like "returns 0 duration when ...".

controllers/loadtest_controller_test.go
```diff
@@ -759,3 +762,87 @@ var _ = Describe("Pod Creation", func() {
 		})
 	})
 })
+
+var _ = Describe("getRequeueTime", func() {
+	var test *grpcv1.LoadTest
+	var reconciler *LoadTestReconciler
+
+	BeforeEach(func() {
```
codeblooded, line 770: Great job with tests. You are a native ginkgo/BDD speaker now.

codeblooded (commented):
I think this is nearing ideal state. I have a few minor comments and a question that I don't have the answer to. It may be sufficient to not address that question for the moment and follow-up with an enhancement to address it.

controllers/loadtest_controller.go
wanlin31: I remembered that :) Will fix them.

controllers/loadtest_controller.go
wanlin31: I feel there is not any convenient way to deal with re queue of the event.

The worst case scenario:
1. Not re queue with timeout --> not able to catch it when it did went over time -->  the cleanup agent won't know about this --> would not stop the workers if they are still running --> will depend on TTL to delete them.
This worse case scenario would cause some resource waste, but if it happened too frequently, user interference might be required to resolved it. 

2. Not re queue to see it at TTL --> the load test will live on the cluster possibly forever --> possibility causing resource waste, this again would require human attention if this happened a lot. 

I agree it would be nice to have something that independent,  just do a cleanup up for everything on the cluster (delete the ones live out of TTL), I also agree this is more like a improvement later, if we do need this, it should not be too difficult to add down the road.

controllers/loadtest_controller_test.go
wanlin31: Right, I thought the conditions has to be in the context...I will modify it.

config/samples/cxx_example_loadtest.yaml
wanlin31: I definitely should change this since it has to be in agreement with states names in Loadtest.LoadTestState.

controllers/loadtest_controller.go
```diff
@@ -139,12 +155,39 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		}
 	}
 
+	reQueueTime := getReQueueTime(test, previousStatus, log)
```
codeblooded, line 158: I hope this doesn't sound difficult, but I prefer the `requeueTime` and `getRequeueTime` casing, since it matches `ctrl.Result{Requeue: true}` and `ctrl.Result{RequeueAfter: ...}`.

codeblooded (approved):
LGTM after 1 fix and @paulosjca may have comments on wording for documentation. 

Let's hold off on merging until after #92 and a deployment, since they most certainly will conflict.

controllers/loadtest_controller.go
wanlin31: No, you are helping find things I missed, I appreciate that! I fix it after the meeting.:) I also like to keep it tidy.

config/samples/cxx_example_loadtest.yaml
```diff
@@ -50,6 +50,21 @@ spec:
   # results:
   #   bigQueryTable: "example-project.foo.demo_dataset"
 
+  # timeoutSeconds field is to advise the longest time a test
```
paulosjca, line 53: I would rewrite this as follows:

```yaml
  # timeoutSeconds is an integer field that indicates the longest time a test
  # is allowed to run, in seconds. Tests that run longer than the given value
  # will be marked as Errored and will no longer be allocated resources to run.
  # For example: timeoutSeconds: 900 indicates the timeout of this test
  # is 15min. The minimum valid value for this field is 1.
```

config/samples/cxx_example_loadtest.yaml
```diff
@@ -50,6 +50,21 @@ spec:
...
+  # could possibly run, test runs longer than given value will be marked as
+  # Errored and would not be allocate more resource. The field takes integer
+  # and convert the supplied value to duration, the unit of duration is fixed as
+  # second, for example: timeoutSeconds: 900 indicates the timeout of this test
+  # is 15min. The minimum seconds one can put is 1.
+  timeoutSeconds: 900
+
+  # ttlSeconds field is to advise on how long the test should be alive on the
```
paulosjca, line 61: I would rewrite this as follows:

```yaml
  # ttlSeconds is an integer field that indicates how long a test is allowed to
  # live on the cluster, in seconds. Tests that live longer than the given value
  # will be deleted. For example: ttlSeconds: 86400 indicates the time-to-live
  # of this test is 24h. The minimum valid value for this field is 1.
```

config/samples/go_example_loadtest.yaml
```diff
@@ -50,6 +50,21 @@ spec:
   # results:
   #   bigQueryTable: "example-project.foo.demo_dataset"
 
+  # timeoutSeconds field is to advise the longest time a test
```
paulosjca, line 53: The C++ comments apply to the other languages as well.

controllers/loadtest_controller.go
```diff
@@ -196,12 +212,39 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	if requeueTime != 0 {
+		return ctrl.Result{RequeueAfter: requeueTime}, nil
+	}
 	return ctrl.Result{}, nil
 }
 
-// LoadTestMissing categorize missing components based on their roles at specific
-// moment. The struct is a wrapper to help us get role information associate
-// with components.
+// getRequeueTime takes a LoadTest and its previous state, and compares the
```
paulosjca, line 222: I would rewrite this as follows:

```go
// getRequeueTime takes a LoadTest and its previous status, compares the
// previous status of the load test with its updated status, and returns a
// calculated requeue time. If the test has just been assigned a start time
// (i.e., it has just started), the requeue time is set to the timeout value
// specified in the LoadTest. If the test has just been assigned a stop time
// (i.e., it has just terminated), the requeue time is set to the time-to-live
// specified in the LoadTest, minus its actual running time. In other cases,
// the requeue time is set to zero.
```

config/samples/cxx_example_loadtest.yaml
wanlin31: This is much more clearer! Thank you!

controllers/loadtest_controller.go
wanlin31: This is great, thank you for reviewing! I forgot to describe what the default value is. I would describe what the function does in all the cases next time, this is great to learn.

api/v1/loadtest_types.go
```diff
@@ -264,9 +264,9 @@ var FailedSettingDefaultsError = "FailedSettingDefaults"
 // in the Initializing state.
 var PodsMissing = "PodsMissing"
 
-// TimeoutExceeded is the reason string when the load test has not yet terminated
+// TimeoutErrored is the reason string when the load test has not yet terminated
 // but exceeded the timeout.
-var TimeoutExceeded = "TimeoutExceeded"
+var TimeoutErrored = "imeoutErrored"
```
paulosjca, line 269: There is a typo here. I thought TimeoutExceeded was fine, but it is ok to change to TimeoutErrored as well.

wanlin31: Right, I think it is a type of error reason, so it made scene to have the 'error' word in the reason itself.

paulosjca (approved):

PR #92, opened 2020-11-16
Embed Scenarios in load tests

codeblooded:
This change adds support for embedding a JSON string with the Scenarios message on a load test. Under the hood, the controller will create and manage the ConfigMap. It also removes support for referencing a manually created ConfigMap. In accordance with this change, this commit also modifies all of the examples to support this new style.

Finally, it comments out some tests with TODO markers. In a subsequent commit, the tests will be improved and cover this embedding thoroughly.


controllers/suite_test.go
```diff
@@ -236,9 +236,7 @@ func newLoadTest() *grpcv1.LoadTest {
 				BigQueryTable: &bigQueryTable,
 			},
 
-			Scenarios: []grpcv1.Scenario{
-				{Name: "cpp-example-scenario"},
-			},
+			ScenariosJSON: "{\"name\": \"cpp-example-scenario\"}",
```
jtattermusch, line 239: it feels odd that you're assigning something that actually doesn't have the prescribed format.
the top-level `{ "scenarios": [` seems missing.
Also below.

config/crd/bases/e2etest.grpc.io_loadtests.yaml
```diff
@@ -781,20 +781,11 @@ spec:
...
-                  name:
-                    description: Name identifies the name of the ConfigMap with the
-                      scenario data.
-                    type: string
-                required:
-                - name
-                type: object
-              type: array
+            scenariosJSON:
+              description: 'ScenariosJSON is string with the contents of a Scenarios
```
jtattermusch, line 785: I think at this point we are only supporting the setup where the top level "scenarios" list has a single item. You should document the current and intended state somewhere. Looks like we might NOT want to support multiple scenarios in a single JSON because the json is passed as-is to the driver? Or maybe that's desirable? Not sure, but the behavior should be documented.

jtattermusch (commented):
+1 to simplifying the workflow. The general idea is looking good, I'll let other to review in detail, I left a few comments though.

controllers/suite_test.go
codeblooded: @jtattermusch I agree 💯%. I want to improve the testing suite in general. I will likely follow this PR up with changes for the test suite and better tests, because it will be a substantially large change.

config/crd/bases/e2etest.grpc.io_loadtests.yaml
codeblooded: The driver expects the flag `--scenarios_file` or `--scenarios_json`. Both of these are the Scenarios message, which supports multiple scenarios. We decided to leave it this way, because we did not want to unmarshal, add and marshal the JSON once to add the appropriate `{"scenarios": [ ... ]}` wrapper.

I felt that continuing to support multiple scenarios in the same containers may actually serve as a benefit for advanced usage and debugging. For example, I may schedule a Java scenario and then run a second. I may also run the second alone. There may be cases were hermetic testing hides issues. If I see a degradation in performance in comparison to running it alone, it suggests there may be some leak.

All of our examples will continue to show 1 scenario per load test, and this is how our CI will run for now. However, we left the `scenariosJSON` plural in the case it may be beneficial later.

codeblooded:
Marking Ready, since the tests will be improved in a follow up PR, and this is working as expected.

config/crd/bases/e2etest.grpc.io_loadtests.yaml
jtattermusch: I agree that continuing to support whatever the driver accepts (=multiple scenarios in a single json) makes sense - I just think it might be useful to document that - users won't understand the details of the implementation and thus they won't know whether they should pass multiple scenarios in a single json or embed multiple scenarios (they woudn't even know what the difference between the two options is). So the docs should say something like "while multiple scenarios in a single JSON is supported, for better isolation it is recommended to specify multiple entries of ScenariosJSON because XYZ".

config/samples/go_example_loadtest.yaml
```diff
@@ -50,14 +50,67 @@ spec:
...
-  scenarios:
-    - name: go-example-scenario
+  # ScenariosJSON is string with the contents of a Scenarios message, formatted
+  # as JSON. See the Scenarios protobuf definition for details:
+  # https://github.com/grpc/grpc-proto/blob/master/grpc/testing/control.proto.
+  scenariosJSON: |
+    {
+      "scenarios": [
+        {
+          "name": "go_example_scenario",
```
jtattermusch, line 60: nit: the C++ embeded example has a name of : ` cpp_generic_async_streaming_ping_pong_secure` - perhaps we want to use the name of the original scenario_config.py scenario here as well?

config/samples/java_example_loadtest.yaml
```diff
@@ -62,14 +62,70 @@ spec:
...
-  scenarios:
-    - name: java-example-scenario
+  # ScenariosJSON is string with the contents of a Scenarios message, formatted
+  # as JSON. See the Scenarios protobuf definition for details:
+  # https://github.com/grpc/grpc-proto/blob/master/grpc/testing/control.proto.
+  scenariosJSON: |
+    {
+      "scenarios": [
+        {
+          "name": "java_example_scenario",
```
jtattermusch, line 72: nit: use the "real" name of the scenario? (same as with Go example)

config/samples/python_example_loadtest.yaml
```diff
@@ -49,15 +49,73 @@ spec:
...
-    - name: python-example-scenario
+  
+  # ScenariosJSON is string with the contents of a Scenarios message, formatted
+  # as JSON. See the Scenarios protobuf definition for details:
+  # https://github.com/grpc/grpc-proto/blob/master/grpc/testing/control.proto.
+  scenariosJSON: |
+    {
+      "scenarios": [
+        {
+          "name": "python_example_scenario",
```
jtattermusch, line 60: nit: use the "real" name of the scenario? (same as with Go example)

config/samples/cxx_example_loadtest.yaml
```diff
@@ -50,14 +50,70 @@ spec:
...
-  scenarios:
-    - name: cxx-example-scenario
+  # ScenariosJSON is string with the contents of a Scenarios message, formatted
+  # as JSON. See the Scenarios protobuf definition for details:
+  # https://github.com/grpc/grpc-proto/blob/master/grpc/testing/control.proto.
+  scenariosJSON: |
+    {
+      "scenarios": [
+        {
+          "name": "cpp_generic_async_streaming_ping_pong_secure",
```
jtattermusch, line 60: nit: add a note that the exact contents of this JSON were generated by the scenario_config.py file?

jtattermusch: qq: what will be the name of the scenario that will end up in bigQuery if I choose to upload the results? Will it be "cpp_generic_async_streaming_ping_pong_secure"? (that's what I would expect and also AFAIR that's what the legacy system does). 
I think there should be a comment that mentions that (if seems quite important).

codeblooded: Yes, the scenario name will be used in BigQuery, since uploads is a responsibility of the driver.

config/crd/bases/e2etest.grpc.io_loadtests.yaml
codeblooded: We have modified the internal user guide to reflect this. Changes are not yet merged.

paulosjca (approved):

PR #97, opened 2020-12-09
Refactor pod creation logic and add some e2e tests

codeblooded:
This commit removes the Component type and adds some e2e tests.

The Driver, Server and Client types shared fields through composition with a Component. However, this made the code hard to understand. It also made it difficult to add a field to one of these types but not all of them. By removing the Component type and composition, the code is simplified and more future-proof. In addition, the first commit moves the logic for creating pods out of the controller package. It adds a new podbuilder package, which contains a PodBuilder type. This type can create pods for Drivers, Servers and Clients.

The controller lacked integration / e2e tests with a pseudo-kubernetes environment. This change introduces some tests using controller-runtime's envtest.  Additional tests will likely be added in follow-up PRs; perhaps, with bug fixes.


controllers/loadtest_controller_test.go
```diff
@@ -1,763 +1,315 @@
...
-				Value: gitRef,
-			}))
-		})
+			return fetchedTest.Status.State, nil
+		}).Should(Equal(grpcv1.Errored))
 	})
 
-	Describe("newBuildContainer", func() {
-		var build *grpcv1.Build
+	It("updates the status was pods are running", func() {
```
codeblooded, line 220:
```suggestion
	It("updates the status when pods are running", func() {
```

podbuilder/podbuilder.go
```diff
@@ -0,0 +1,351 @@
...
+			Name:      fmt.Sprintf("%s-%s-%s", pb.test.Name, pb.role, pb.name),
+			Namespace: pb.test.Namespace,
+			Labels: map[string]string{
+				config.LoadTestLabel:      pb.test.Name,
+				config.RoleLabel:          pb.role,
+				config.ComponentNameLabel: pb.name,
+			},
+		},
+		Spec: corev1.PodSpec{
+			NodeSelector: map[string]string{
```
paulosjca, line 287: I believe that the logic here could be enhanced to add the "pool" only if it is not empty, and if it is empty, add a "default_pool" corresponding to the pb.role. So the controller configuration could (optionally) specify something like default_worker_pool and default_driver_pool. and that would be added to the "default_pool(s)" map. It would be up to the maintainer of the cluster to ensure that nodes in the cluster are labeled, say, "default_pool=worker", in addition to, say, "pool=worker-8core", in order to make use of this feature.

config/defaults.go
```diff
@@ -119,66 +119,48 @@ func (d *Defaults) Validate() error {
 // system cannot infer a run image for "fortran" if a build image was not
```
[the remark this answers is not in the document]
codeblooded: That is the current behavior. An unknown language must specify images.

podbuilder/podbuilder.go
codeblooded: @paulosjca But these default pools should not live in the config/defaults.yaml file, right? After our discussion, would it make sense to have these default pools set in GCP Console where they are maintained? Then, it will try to use those pools. The only problem with defaults is that we do not know if they exist beforehand. It will most likely not produce a clean error, but try repeatedly in the pending state.

paulosjca: The enhancement here can be added in a follow-up PR.

codeblooded:
Converting to draft to fix. Somehow the rebasing messed things I up. Will mark ready once fixed.

codeblooded:
The commits were really butchered. Some state was lost. I did not think it was worth it to hunt in the reflog, since both of these changes are non-functional. Therefore, I squashed them with a reasonable commit message.

paulosjca (approved):

podbuilder/podbuilder.go
```diff
@@ -0,0 +1,351 @@
...
+	if defs == nil {
+		return corev1.Container{}
+	}
+
+	var args []string
+	for _, server := range test.Spec.Servers {
+		args = append(args, fmt.Sprintf("%s=%s,%s=%s,%s=%s",
+			config.LoadTestLabel, test.Name,
+			config.RoleLabel, config.ServerRole,
+			config.ComponentNameLabel, *server.Name,
```
wanlin31, line 73: Nit: Wondering if we are trying to eliminate the component object why are we still using `componentNameLabel` ? If it means different from what it was you might want to change the comment for `componentNameLabel` in constant.go.

config/defaults_test.go
```diff
@@ -1,664 +1,664 @@
-/*
-Copyright 2020 gRPC authors.
+// /*
```
wanlin31, line 1: Little confused about what is happening with this file, comment all of them out? If so, we would not have any unite test for defaults.go.

codeblooded: Good catch!

podbuilder/podbuilder.go
codeblooded: Agreed. Will figure out a rename or reword.

controllers/loadtest_controller_test.go
```diff
@@ -1,850 +1,309 @@
...
-				Command: []string{"go"},
-				Args:    []string{"run", "main.go"},
+	It("does not change the status after termination", func() {
+		now := metav1.Now()
+		test.Status = grpcv1.LoadTestStatus{
+			State:     grpcv1.Succeeded,
+			StartTime: &now,
+			StopTime:  &now,
+		}
+		Expect(k8sClient.Create(context.Background(), test)).To(Succeed())
```
wanlin31, line 75: Wondering if there is any difference between using Expect() or Eventually()

controllers/loadtest_controller_test.go
```diff
@@ -1,850 +1,309 @@
...
+				}
 			}
 
-			run.Env = env
-
-			container := newRunContainer(run)
-			Expect(env[0]).To(BeElementOf(container.Env))
-			Expect(env[1]).To(BeElementOf(container.Env))
-		})
+			return foundPodCount, nil
```
wanlin31, line 158: If the function returns 2 values and the expectedPodCount is just 1 value, how would they match? Will the nil not be counted?

controllers/loadtest_controller_test.go
```diff
@@ -1,850 +1,309 @@
...
-
-var _ = Describe("getRequeueTime", func() {
-	var test *grpcv1.LoadTest
-	var reconciler *LoadTestReconciler
-
-	BeforeEach(func() {
-		test = newLoadTest()
-		reconciler = &LoadTestReconciler{
-			Log: ctrl.Log.WithName("controllers").WithName("LoadTest"),
+	It("updates the status when pods terminate with errors", func() {
```
wanlin31, line 162: "updates the status when pods terminate with errors", means update the status of the test I assume, is that right?

controllers/loadtest_controller_test.go
```diff
@@ -1,850 +1,309 @@
...
-		It("returns the test's timeout", func() {
-			var startTime metav1.Time
-			expectedRequeueTime := time.Duration(test.Spec.TimeoutSeconds) * time.Second
-			previousStatus := grpcv1.LoadTestStatus{}
+		}
+		if testSpec.Driver != nil {
+			pod = builder.PodForDriver(testSpec.Driver)
+			Expect(createPod(pod, test)).To(Succeed())
+			Expect(updatePodWithContainerState(pod, errorState)).To(Succeed())
+		}
```
wanlin31, line 187: I think this portion of the code is saying any one of the pod that went errored will cause the entire LoadTest to be marked as errored, so we only need one errored pod, why are we creating every pods thats needed in the spec?

controllers/loadtest_controller_test.go
codeblooded: Yes.

wanlin31:
This looks great to me, just have some questions regarding tests.:)

controllers/loadtest_controller_test.go
wanlin31: Nit: Wondering if we should add "test" in the it clause, just clearer to me. :)

config/defaults.go
```diff
@@ -119,66 +119,54 @@ func (d *Defaults) Validate() error {
...
-
-	if component.Name == nil {
-		name := uuid.New().String()
-		component.Name = &name
+// setNameOrDefault replaces the address of its pointer argument with the
+// address of a UUID string. If the argument is nil upon invocation, it will
+// remain nil.
+func (d *Defaults) setNameOrDefault(namePtr *string) {
+	if namePtr != nil {
+		return
```
wanlin31, line 153: A bit confused about this function, the function itself is well self explanatory: if we namePtr is not nil we do not anything, and if it is nil we give it a UUID. This seems not consistent with the comment, could you help me understand this?

podbuilder/podbuilder.go
codeblooded: @wanlin31 I'm not sure of a better name for this. I don't want to call it the pod name b/c it is not. I think drivers, servers and clients are still components of the test. So, we may just want to leave as is for the moment and revisit if we have time later.

config/defaults.go
codeblooded: @wanlin31 Sorry, I pushed revision 5e6d98a. Let me know if that fixes it.

wanlin31 (approved):

codeblooded:
@wanlin31 how is bb80739? Changed setNameOrDefault to unwrapStrOrUUID. The accidentally commented out tests caught two bugs. This change fixes them.

wanlin31:
Looks good to me!

PR #98, opened 2020-12-11
Feature/csharp image

wanlin31:
This commit updates the example load tests configuration yaml file.


wanlin31:
I realized that the worker was built by run_test.py (automatically go through pre-build, build and post-build process), more convenient than building the workers manually. We need python installed if we want to run run_test.py, in the original commit it was a python 2.7, would try to see if python3 works too.

The run image is updated to docker official dotnet image and verified with tests.

wanlin31:
I just think we would probably want to use a dotnet official image to build.

wanlin31:
Just remembered that IMAGE_PREFIX=gcr.io/grpc-testing/e2etest/runtime/ , might need to figure out a way to alter this since this time the customized image is a build image.

Makefile
```diff
@@ -139,6 +139,14 @@ python-image:
...
 	docker push ${IMAGE_PREFIX}python:${TEST_INFRA_VERSION}
 
+# Build the Csharp build image
+csharp-image:
+	docker build -t ${INIT_IMAGE_PREFIX}build/csharp:${TEST_INFRA_VERSION} containers/init/build/csharp
+
+# Push the cshsarp build image to a docker registry
+push-python-image:
+  docker push ${IMAGE_PREFIX}csharp:${TEST_INFRA_VERSION}
+
```
codeblooded, line 149: I think there's a typo here. 

```suggestion
# Push the C# build image to a docker registry
push-csharp-image:
  docker push ${INIT_IMAGE_PREFIX}build/csharp:${TEST_INFRA_VERSION}

```

We may want to create a ${BUILD_IMAGE_PREFIX} = ${INIT_IMAGE_PREFIX}/build/ environment variable at the top. That may be a little cleaner and less prone to errors. What do you think?

containers/init/build/csharp/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN apt-get update && apt-get install -y cmake && apt-get clean
+
+# Python is needed to run the run_tests.py script (we use it to build C#)
+RUN apt-get update && apt-get install -y \
+  python-dev \
+  python-setuptools \
+  python-yaml && \
+  apt-get clean
+
+RUN curl https://bootstrap.pypa.io/get-pip.py | python2.7
```
codeblooded, line 38: If we stick with run_tests.py... I thought we decided not to use python2.7? If you are having trouble getting python3 to work with grpc tools, ping me... I've been through the problem before. 

While we do not have to use run_tests.py, it is the easiest according to https://github.com/grpc/grpc/tree/master/src/csharp#build-from-source. There are other ways which would remove the need for python. But looking at the code for run_tests.py, C# looks a little complicated https://github.com/grpc/grpc/blob/9293ce3e81d5e6a3fc07e3263210f159e3b5b9cc/tools/run_tests/run_tests.py#L903.

codeblooded (commented):
A few comments, but LGTM. We may not want to include all of these example scenarios. We may just want to include one like the other languages. Then, we could use the same naming convention as well.

containers/init/build/csharp/Dockerfile
wanlin31: I haven't update this yet, I would eliminate the run_test.py as we discussed in the meeting, and therefor we would no longer need python for anything. :)

wanlin31:
This is still a work in progress, as we discussed, we are not going to use run_test.py since it adds an uncessary layer, would update the build image to reflect this change.

wanlin31:
I left my personal image registry in examples just in case you want to try the examples, will be removed later.

wanlin31:
I have updated the pr to ready to merge state. The images were verified by running through the deployment process on the staging cluster.

Makefile
```diff
@@ -163,6 +163,14 @@ python-image:
 push-python-image:
 	docker push ${IMAGE_PREFIX}python:${TEST_INFRA_VERSION}
 
+# Build the Csharp build image
```
paulosjca, line 166: I would write "csharp" for consistency.

paulosjca (approved):

config/defaults_template.yaml
```diff
@@ -31,3 +31,6 @@ languages:
     buildImage: l.gcr.io/google/bazel:latest
     runImage: "{{ .ImagePrefix }}python:{{ .Version }}"
 
+  - language: csharp
+    buildImage: "{{ .InitImagePrefix }}build/csharp:{{ .Version }}"
```
codeblooded, line 35: We may want to create a new variable, `.BuildImagePrefix` which is the `.InitImagePrefix` + `"build/"` or some override.

The problem is that the image "build/csharp" may not be valid for all registries. DockerHub, for instance, would not likely work. With the other examples, the images are valid even if the `.InitImagePrefix` or `.ImagePrefix` are not set. 

We don't have to do it in this PR, since we know we are only using gcr at the moment. But, we should do it... since we have avoided most vendor lock-in thus far.

config/samples/csharp_example_loadtest.yaml
```diff
@@ -0,0 +1,129 @@
...
+  # is an array for future expansion.
+  #
+  # There are many designs and systems to pursue load balancing,
+  # organizing and monitoring a mesh of servers. Therefore, this
+  # will likely be expanded in the future.
+  servers:
+    - language: csharp
+      clone:
+        repo: https://github.com/jtattermusch/grpc.git
+        # TODO(jtattermusch): we cannot use grpc/grpc master because currently C# qps_worker doesn't accept the --server_port arg
```
codeblooded, line 27: We removed server ports in #97 refactor, they are completely scrubbed from the defaults in #99. It may be better to wait for one of those, so we can use the official repo here.

wanlin31: I think this PR is mostly about adding support for csharp, example is just a part of it. Since the main part is done, I think it may be better merge this first and update it when 97 get merged. I am okay with both solution actually. :)

Makefile
```diff
@@ -163,6 +163,14 @@ python-image:
 push-python-image:
 	docker push ${IMAGE_PREFIX}python:${TEST_INFRA_VERSION}
 
+# Build the csharp build image
+csharp-image:
+	docker build -t ${INIT_IMAGE_PREFIX}build/csharp:${TEST_INFRA_VERSION} containers/init/build/csharp
+
+# Push the csharp build image to a docker registry
+push-csharp-image:
+	docker push ${INIT_IMAGE_PREFIX}build/csharp:${TEST_INFRA_VERSION}
```
codeblooded, line 172: Similar to the configure script where we introduced a new variable for BUILD_IMAGE_PREFIX, I think we should do that in the Makefile. If I set $INIT_IMAGE_PREFIX="" and try to push to DockerHub registry, it would push as build/csharp which would not be valid. So, I'd like to move the build/ part to be customizable as well. Can we introduce a ${BUILD_IMAGE_PREFIX}? It would also require a change to deployment guide.

codeblooded (approved):
LGTM just one nit ask of Makefile

Makefile
wanlin31: If we have this BUILD_IMAGE_PREFIX, should we also ask BUILD_IMAGE_PREFIX to be set up during the deployment process? If we do, since build is part of the init containers we separating them with two environment variables could create some confusion. I agree we should make sure it is compatible with all registries, just feel the naming of the environment is a bit confusing.

containers/init/build/csharp/build_qps_worker.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+bash /src/workspace/tools/run_tests/helper_scripts/pre_build_csharp.sh
```
jtattermusch, line 18: this script is considered to be part of run_tests.py's internals/implementation detail so I wouldn't rely on its exact semantics here (btw, you're also setting MSBUILD_CONFIG and CONFIG env variables in the Dockerfile to match the semantics, which feels hacky)

I think it's better to just "inline" the interesting contents of "pre_build_csharp.sh" since the only interesting command is the cmake command: https://github.com/grpc/grpc/blob/e745f8cb558798f1c8b91baa42d46228a9d4a0c4/tools/run_tests/helper_scripts/pre_build_csharp.sh#L24

Basically, just do this (it's more consistent, readable and also simpler)
```
# Build C# native extension with cmake
mkdir -p cmake/build
pushd cmake/build
cmake -DgRPC_BUILD_TESTS=OFF -DCMAKE_BUILD_TYPE=Release ../..
make grpc_csharp_ext
popd

cd src/csharp
dotnet publish Grpc.IntegrationTesting.QpsWorker/ -c Release -f netcoreapp2.1 -o ../../../qps_worker
```

containers/init/build/csharp/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+  git && \
+  apt-get clean
+
+# The prerequisites installed below are heavily inspired by the C# testing image from grpc/grpc
+# https://github.com/grpc/grpc/blob/master/tools/dockerfile/test/csharp_stretch_x64/Dockerfile
+
+# cmake >=3.6 needed to build grpc_csharp_ext
+RUN apt-get update && apt-get install -y cmake && apt-get clean
+
+ENV MSBUILD_CONFIG=Release
```
jtattermusch, line 31: let's remove these env variables, you don't really need them and they obscure the build process. See comment below for more details.

jtattermusch (commented):
Reviewed from the gRPC C# perspective. It's looking mostly good but I have some comments regarding the Dockerfile and the build script.

containers/init/build/csharp/build_qps_worker.sh
wanlin31: Agreed, this make the Dockerfile  much clearer, thank you for reviewing. I will make the change accordingly.

containers/init/build/csharp/Dockerfile
wanlin31: Good to learn! Since we are not using the pre_build_csharp.sh anymore, there is not need to set up these variables. Will remove them.

jtattermusch (approved):
LGTM from the C# build perspective.

PR #99, opened 2020-12-11
Remove driverPort and serverPort from defaults

codeblooded:
This change removes the driverPort from the defaults, since we always rely on port 10000. The serverPort is also removed, since Kubernetes permits communication amongst pods on all ports.

### DIFF
--- a/config/constants.go
+++ b/config/constants.go
@@ -58,8 +58,9 @@ const (
 	// on a driver component.
 	DriverRole = "driver"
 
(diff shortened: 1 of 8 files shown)


paulosjca (approved):

wanlin31:
It doesn't let me  add myself either. I will read it and ask you if I have any questions. :)

wanlin31 (approved):
LGTM!

PR #100, opened 2020-12-11
Remove default pool options

codeblooded:
This change removes default pool options, requiring a pool to be specified on each test driver, server and client.

 **NOTE**: This change uses #97; therefore it should be reviewed and merged first. Then, this will be rebased.


codeblooded:
Closing to implement a different way. Will reference this original PR in the new and improved one.

PR #101, opened 2020-12-30
Replace default pools with default pool labels

codeblooded:
This commit removes the requirement for the names of default pools to be specified in the defaults configuration. Instead, it adds the option to specify the name of a default pool label for a client, driver and server. Therefore, the workflow becomes:

 - If the user has specified a pool for a client, driver or server; its pod will be scheduled on a node with a matching value for the "pool" label.

 - If the user has not specified a pool but the defaults configuration specifies default pool labels, the pod will be scheduled on a node with the appropriate label present. For example, consider a LoadTest with a client that has no pool specified. The defaults configuration specifies "default-pool-client" as the string for the client field on the defaultPoolLabels struct. The client's pod will be scheduled on any node with a label named "default-pool-client" and a string value of "true".

 - If the user does not specify a pool and there are no default pool labels, the LoadTest will error.

### DIFF
--- a/config/defaults_template.yaml
+++ b/config/defaults_template.yaml
@@ -1,8 +1,9 @@
 # See ./defaults.go for documentation on each field.
 
-driverPool: "{{ .DriverPool }}"
-
-workerPool: "{{ .WorkerPool }}"
+defaultPoolLabels:
+  client: default-client-pool
+  driver: default-driver-pool
+  server: default-server-pool
 
 cloneImage: "{{ .InitImagePrefix }}clone:{{ .Version }}"
 
--- a/config/defaults_test.go
+++ b/config/defaults_test.go
@@ -29,11 +29,14 @@ var _ = Describe("Defaults", func() {
 	BeforeEach(func() {
 		defaults = &Defaults{
 			ComponentNamespace: "component-default",
-			DriverPool:         "drivers",
-			WorkerPool:         "workers-8core",
-			CloneImage:         "gcr.io/grpc-fake-project/test-infra/clone",
-			ReadyImage:         "gcr.io/grpc-fake-project/test-infra/ready",
-			DriverImage:        "gcr.io/grpc-fake-project/test-infra/driver",
+			DefaultPoolLabels: &PoolLabelMap{
+				Client: "default-client-pool",
+				Driver: "default-driver-pool",
+				Server: "default-server-pool",
+			},
+			CloneImage:  "gcr.io/grpc-fake-project/test-infra/clone",
+			ReadyImage:  "gcr.io/grpc-fake-project/test-infra/ready",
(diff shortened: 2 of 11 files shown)


config/defaults_template.yaml
```diff
@@ -1,8 +1,9 @@
 # See ./defaults.go for documentation on each field.
 
-driverPool: "{{ .DriverPool }}"
-
-workerPool: "{{ .WorkerPool }}"
+defaultPoolLabels:
+  client: default-client-pool
+  driver: default-driver-pool
+  server: default-server-pool
```
codeblooded, line 6: I think it may be better to make these like LabelSelectors. Instead of relying on a hardcoded "true" string... maybe we should make this specify both a key and a value. wdyt?

config/defaults_test.go
```diff
@@ -173,14 +164,6 @@ var _ = Describe("Defaults", func() {
 				Expect(driver.Name).ToNot(BeNil())
 			})
 
-			It("sets default pool when unspecified", func() {
```
codeblooded, line 176: /todo Add a couple tests here before merging.

codeblooded:
@grpc/test-infra I have a couple more things I would like to do here. However, I'm opening for review to start a discussion on these changes.

config/defaults_template.yaml
paulosjca: Reading the documentation, that makes sense:

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

So, if the pool is not specified, a label selector is added instead.

One concern I would have is that the implementation might be more complex, but perhaps not.

If it takes time to do, it can be done in a follow-up PR, since it does not change the interface.

paulosjca: I misread the documentation in the comment above. The label always has a value, so the most we should do is make the value a constant in our code, instead of hardcoding to "true" (since it appears in three places).

We should not change the interface to require the user to specify a value, because that would make life more difficult for the user.

wanlin31, line 6: If we already have the default for driver and workers, do we still want the configuration.go file to error out when the pool is not set? See https://github.com/grpc/test-infra/blob/23ce525193460dbd1633a080e85b72cacd1e0784/config/cmd/configure.go#L104-L110

codeblooded: Oops. How did that get missed? I'll add it.

wanlin31: Also if "default-client-pool", "default-driver-pool" and "default-server-pool" become our keys for our NodeSelectors keys, the values corresponding to those strings would be labels of the Nodes, wondering how do we know what are the labels that the Node are having? Confused....

codeblooded: Let's talk about it over gvc. 👍🏻

paulosjca (approved):

wanlin31 (approved):

PR #102, opened 2021-01-04
Rename cleanup agent test suite

codeblooded:
Let's rename the cleanup test suite to **Cleanup Agent Suite** in order to distinguish it from the **Controller Suite**.


wanlin31 (approved):
Good catch!

PR #104, opened 2021-01-14
Gang Scheduling

codeblooded:
The default Kubernetes scheduler considers each pod individually when scheduling. The only relationships it understands between pods are constraints regarding coexistence on nodes. Unfortunately, this is not sufficient for gRPC load tests. Client pods require server pods in order to send traffic. Both client pods and server pods require a driver pod in order to receive their instructions. Therefore, all of a test's pods must be alive on the cluster for the test to start and complete.

The number of tests that can run is limited by the number of nodes available in the pools the test requires. If a node becomes available for a pod, the default scheduler will immediately schedule it without regard to whether the other pods required for the test are scheduled. This has led to a starvation problem where the cluster becomes filled with some but not all pods required for any test to complete.

Requiring the scheduler to assign related pods to nodes together is referred to as gang scheduling. It is not implemented in Kubernetes. This commit implements a naive gang scheduling system by delaying the creation of a pod for a load test when the cluster cannot accommodate all of the load test's pods.

This pull request adds support for gang scheduling. Gang scheduling means that the load test is considered as an atomic unit when making a decision to schedule any of its pods. If the cluster's current node availability cannot accommodate all of the test's pods, the controller will now defer scheduling any of them. When nodes become available to accommodate all of the test's pods, all pods are scheduled at the same time.

Note this pull request diverges from best practice for a custom Kubernetes controller. Scheduling logic does not belong in a controller. However, the Kubernetes scheduler is complex and has many factors that are considered when making scheduling decisions. Implementing a naive alternative scheduler would be costly. In addition, a controller is expected to be level-driven and operate with a cache. Scheduling decisions require a current view of the cluster; therefore, these changes manually invalidate the cache. This create/delay approach is a compromise due to the current limitations of Kubernetes.

### DIFF
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -79,8 +79,8 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	rawTest := new(grpcv1.LoadTest)
 	if err = r.Get(ctx, req.NamespacedName, rawTest); err != nil {
 		log.Error(err, "failed to get test", "name", req.NamespacedName)
-		// do not requeue, the test may have been deleted or the cache is invalid
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		err = client.IgnoreNotFound(err)
+		return ctrl.Result{Requeue: err != nil}, err
 	}
 
 	testTTL := time.Duration(rawTest.Spec.TTLSeconds) * time.Second
@@ -98,7 +98,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 				return ctrl.Result{Requeue: true}, err
 			}
 		}
-		return ctrl.Result{}, nil
+		return ctrl.Result{Requeue: false}, nil
 	}
 
 	// TODO(codeblooded): Consider moving this to a mutating webhook
@@ -111,12 +111,12 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		if err = r.Status().Update(ctx, test); err != nil {
 			log.Error(err, "failed to update test status when setting defaults failed")
 		}
-		return ctrl.Result{}, err
+		return ctrl.Result{Requeue: false}, nil
 	}
 	if !reflect.DeepEqual(rawTest, test) {
 		if err = r.Update(ctx, test); err != nil {
 			log.Error(err, "failed to update test with defaults")
-			return ctrl.Result{}, err
+			return ctrl.Result{Requeue: true}, err
 		}
 	}
 
@@ -134,7 +134,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			if updateErr := r.Status().Update(ctx, test); updateErr != nil {
 				log.Error(updateErr, "failed to update status after failure to get scenarios ConfigMap: %v", err)
 			}
-			return ctrl.Result{}, err
+			return ctrl.Result{Requeue: true}, err
 		}
 
 		cfgMap = &corev1.ConfigMap{
@@ -269,7 +269,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 				if updateErr := r.Status().Update(ctx, test); updateErr != nil {
 					log.Error(updateErr, "failed to update status after failure due to requesting nodes from a nonexistent pool")
 				}
-				return ctrl.Result{}, errNonexistentPool
+				return ctrl.Result{Requeue: false}, nil
 			}
 
 			if requiredNodeCount > availableNodeCount {
@@ -282,7 +282,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		createPod := func(pod *corev1.Pod) (*ctrl.Result, error) {
 			if err = ctrl.SetControllerReference(test, pod, r.Scheme); err != nil {
 				log.Error(err, "could not set controller reference on pod, pod will not be garbage collected", "pod", pod)
-				return &ctrl.Result{}, err
+				return &ctrl.Result{Requeue: true}, err
 			}
 
 			if err = r.Create(ctx, pod); err != nil {
@@ -305,7 +305,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 				if updateErr := r.Status().Update(ctx, test); updateErr != nil {
 					logWithServer.Error(updateErr, "failed to update status after failure to construct a pod for server")
 				}
-				return ctrl.Result{}, err
+				return ctrl.Result{Requeue: false}, nil
 			}
 
 			if missingPods.Servers[i].Pool == nil {
@@ -338,7 +338,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 				if updateErr := r.Status().Update(ctx, test); updateErr != nil {
 					logWithClient.Error(updateErr, "failed to update status after failure to construct a pod for client")
 				}
-				return ctrl.Result{}, err
+				return ctrl.Result{Requeue: false}, nil
 			}
 
 			if missingPods.Clients[i].Pool == nil {
@@ -371,7 +371,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 				if updateErr := r.Status().Update(ctx, test); updateErr != nil {
 					logWithDriver.Error(updateErr, "failed to update status after failure to construct a pod for driver")
 				}
-				return ctrl.Result{}, err
+				return ctrl.Result{Requeue: false}, nil
 			}
 
 			if missingPods.Driver.Pool == nil {
@@ -398,7 +398,8 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if requeueTime != 0 {
 		return ctrl.Result{RequeueAfter: requeueTime}, nil
 	}
-	return ctrl.Result{}, nil
+
+	return ctrl.Result{Requeue: false}, nil
 }
 
 // getRequeueTime takes a LoadTest and its previous status, compares the
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"time"
@@ -36,9 +37,15 @@ import (
 	"github.com/grpc/test-infra/status"
 )
 
+var (
+	errCacheSync       = errors.New("failed to sync cache")
+	errNonexistentPool = errors.New("pool does not exist")
+)
+
 // LoadTestReconciler reconciles a LoadTest object
 type LoadTestReconciler struct {
 	client.Client
+	mgr      ctrl.Manager
 	Defaults *config.Defaults
 	Log      logr.Logger
 	Scheme   *runtime.Scheme
@@ -50,6 +57,8 @@ type LoadTestReconciler struct {
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=pods/status,verbs=get
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=nodes/status,verbs=get
 
 // Reconcile attempts to bring the current state of the load test into agreement
 // with its declared spec. This may mean provisioning resources, doing nothing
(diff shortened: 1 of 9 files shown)


codeblooded:
Opened kubernetes-sigs/controller-runtime#1360, since I am not currently able to create nodes as necessary to test this change with envtest.

status/missing_pods.go
```diff
@@ -22,28 +22,60 @@ import (
...
 	// indicates the Clients still in need.
-	Clients []grpcv1.Client `json:"clients,omitempty"`
+	Clients []grpcv1.Client
+
+	// NodeCountPyPool is a map which gives the number of nodes required from
+	// each pool to run the test. The names of the required node pools are the
+	// keys and the value is the number of nodes required from the named pool.
+	// If a test does not require any nodes from a pool, the pool name will not
+	// be present as a key in the map.
+	NodeCountByPool map[string]int
```
codeblooded, line 56: TODO: It may be a good idea to add some test cases for default pools. NTS: Need to figure out how to structure them, since they would divert from the Context-blocks that are currently in place. Ideas? @wanlin31

controllers/loadtest_controller.go
```diff
@@ -175,131 +176,221 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+				if updateErr := r.Status().Update(ctx, test); updateErr != nil {
+					logWithDriver.Error(updateErr, "failed to update status after failure to construct a pod for driver")
+				}
+				return ctrl.Result{}, err
+			}
 
-			if updateErr := r.Status().Update(ctx, test); updateErr != nil {
-				logWithDriver.Error(updateErr, "failed to update status after failure to create pod for driver")
+			if missingPods.Driver.Pool == nil {
+				pod.Labels[config.PoolLabel] = defaultDriverPool
```
codeblooded, line 378: It may be a nice enhancement to pass the default pools we calculate into the podbuilder and allow it to set those.

controllers/loadtest_controller.go
```diff
@@ -175,131 +176,221 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-	builder := podbuilder.New(r.Defaults, test)
-
-	createPod := func(pod *corev1.Pod) (*ctrl.Result, error) {
-		if err = ctrl.SetControllerReference(test, pod, r.Scheme); err != nil {
-			log.Error(err, "could not set controller reference on pod, pod will not be garbage collected", "pod", pod)
-			return &ctrl.Result{}, err
+	if !missingPods.IsEmpty() {
+		if !r.mgr.GetCache().WaitForCacheSync(ctx.Done()) {
+			log.Error(errCacheSync, "could not invalidate the cache which is required to gang schedule")
+			return ctrl.Result{Requeue: true}, errCacheSync
```
wanlin31, line 192: Return either Result.{Requeue: true}  or an error would result in a requeue behavior. I am wondering if it is possible to just use Result.{Requeue: true} instead of the error. I am still investigating how the error returned would affect the requeueAfter behavior, seems like there is a connection based on my limited observation.

status/missing_pods.go
```diff
@@ -22,28 +22,60 @@ import (
...
 	// indicates the Servers still in need.
-	Servers []grpcv1.Server `json:"servers,omitempty"`
+	Servers []grpcv1.Server
 
 	// Clients are a list of components that send traffic to servers. The list
 	// indicates the Clients still in need.
-	Clients []grpcv1.Client `json:"clients,omitempty"`
+	Clients []grpcv1.Client
+
+	// NodeCountPyPool is a map which gives the number of nodes required from
```
wanlin31, line 51: NodeCountPyPool --> NodeCountByPool

wanlin31: NodeCountByPool means numbers we still missing or what we need in total? Seems like it is the number that we still missing, I figured this out reading codes below, I think it might be better to update the comment to make it clearer. :)

controllers/loadtest_controller.go
```diff
@@ -175,131 +176,221 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			availableNodeCount, ok := poolAvailabilities[pool]
+			if !ok {
+				log.Error(errNonexistentPool, "requested pool does not exist and cannot be considered when scheduling", "requestedPool", pool)
+				test.Status.State = grpcv1.Errored
+				test.Status.Reason = grpcv1.PoolError
+				test.Status.Message = fmt.Sprintf("requested pool %q does not exist", pool)
+				if updateErr := r.Status().Update(ctx, test); updateErr != nil {
+					log.Error(updateErr, "failed to update status after failure due to requesting nodes from a nonexistent pool")
+				}
+				return ctrl.Result{}, errNonexistentPool
```
wanlin31, line 272: If your intension is not to requeue due to this error you might not want to return this error since whenever an error is return it automatically requeued.

controllers/loadtest_controller.go
```diff
@@ -175,131 +176,221 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			}
 
-		result, err := createPod(pod)
-		if result != nil && !kerrors.IsAlreadyExists(err) {
-			logWithClient.Error(err, "failed to create pod for client")
+			if requiredNodeCount > availableNodeCount {
+				log.Info("cannot schedule test: inadequate availability for pool", "pool", pool, "requiredNodeCount", requiredNodeCount, "availableNodeCount", availableNodeCount)
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
wanlin31, line 277: Based on our current experience, this ctrl.Result{RequeueAfter: 5 * time.Second} might not happen, just a thing to watch out...

controllers/loadtest_controller.go
```diff
@@ -175,131 +176,221 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			pod, err := builder.PodForClient(&missingPods.Clients[i])
+			if err != nil {
+				logWithClient.Error(err, "failed to construct a pod struct for supplied client struct")
+				test.Status.State = grpcv1.Errored
+				test.Status.Reason = grpcv1.ConfigurationError
+				test.Status.Message = fmt.Sprintf("failed to construct a pod for client at index %d: %v", i, err)
+				if updateErr := r.Status().Update(ctx, test); updateErr != nil {
+					logWithClient.Error(updateErr, "failed to update status after failure to construct a pod for client")
+				}
+				return ctrl.Result{}, err
```
wanlin31, line 341: There are many cases you returned an error and ctrl.Result{}, it is a bit confusing if you want to requeue the items or not, we might want to have a rule for what we return such as:
1. want to requeue item: ctrl.Result{Requeue: true}  or error, depend on what you want you could chose either one.
2. do not want to requeue item: ctrl.Result{} + nil
3. want to requeue with a specific time: ctrl.Result{RequeueAfter: Duration} + nil

wanlin31: The above is my naive thought, we could come up with other rules.

controllers/loadtest_controller.go
```diff
@@ -175,131 +176,221 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		for pool, capacity := range poolCapacities {
+			poolAvailabilities[pool] = capacity
+		}
+		for _, pod := range pods.Items {
+			pool, ok := pod.Labels[config.PoolLabel]
+			if !ok {
+				log.Info("encountered a pod without a pool label", "pod", pod)
+				continue
+			}
+			if pod.Status.Phase != corev1.PodSucceeded && pod.Status.Phase != corev1.PodFailed {
```
wanlin31, line 247: Wondering why we used pod.Status.Phase here, as you have advised me earlier it might not be fully updated.

controllers/loadtest_controller.go
wanlin31: I think this is the step you validating cache, wondering how expensive it is

controllers/loadtest_controller.go
codeblooded: Good question. Let me remember what the issue was. I believe it had to do with Pending or Initializing pods. I do not think the status.StateForPod function worked since they had no containers. I'll recheck and update here.

controllers/loadtest_controller.go
codeblooded: I witnessed this only after a failure for >30min to schedule. It is something to lookout for. We can consider two things: 

1. We can adjust the Resync ratio (though not recommended) to re-reconcile everything in the cache at a certain interval as a backup.

1. We could introduce some component in a separate go routine or separate deployment that triggers reconciles after a long period of idle time by bumping the generation number.

controllers/loadtest_controller.go
codeblooded: @wanlin31 It is certainly possible to just use `ctrl.Result{Requeue: true}`, but I do think this is a problematic situation. I do not see a non-exceptional reason why we would not be able to invalidate the cache. I would like to somehow record these errors.

controllers/loadtest_controller.go
codeblooded: Fair point. I still do not understand why `return ctrl.Result{Requeue: false}, errNonexistentPool` does not appear to be an option.

controllers/loadtest_controller.go
codeblooded: For #3, I have not checked controller-runtime again... but I'm guessing the `ctrl.Result{RequeueAfter: ...}` is ignored when the error is also returned?

wanlin31: Yes, if there is a error returned with ctrl.Result{RequeueAfter: ...} then it is requeued same as  ctrl.Result{Requeue: true}

controllers/loadtest_controller.go
wanlin31: well if you do want that, you could log the error, just not to return it.

controllers/loadtest_controller.go
wanlin31: I agree, I think this error is a bit different from "Already exit" and "Update conflict" since the aforementioned two errors are not real errors, one might not want to record them, what we have here could be a real error that needed be recorded.

controllers/loadtest_controller.go
wanlin31: I am reading controller-runtime code base as a hobby, lol, learned a lot. I am suspecting if any reconcile is happening due to ctrl.Result{Requeue: true} or an error returned (not reconcile due to a resource update ) before RequeueAfter time is up, the RequeueAfter could be requested. Still investigating....I think Solly has the point of something is update the lock, although I am not so sure that is what is happening but I think it is a direction worth exploring

controllers/loadtest_controller.go
wanlin31: By recording this error, it affects the rate of retrying

codeblooded: That's a good point.  If it failed to wait for the cache, the reconcile timeout was reached or their was a problem with the API. I don't know. We do need to retry if it fails b/c otherwise the test will never be scheduled.

controllers/loadtest_controller.go
codeblooded: That is crazy. Why can't you have an error with a delay to retry?

codeblooded: @wanlin31 I have addressed the consistency problem in 74fda57. Let me know what you think.

paulosjca (approved):

controllers/loadtest_controller.go
```diff
@@ -111,12 +111,12 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 			log.Error(err, "failed to update test status when setting defaults failed")
 		}
-		return ctrl.Result{}, err
+		return ctrl.Result{Requeue: false}, nil
 	}
 	if !reflect.DeepEqual(rawTest, test) {
 		if err = r.Update(ctx, test); err != nil {
 			log.Error(err, "failed to update test with defaults")
-			return ctrl.Result{}, err
+			return ctrl.Result{Requeue: true}, err
```
wanlin31, line 119: Do you want to requeue this even there is no error? Similar to other ctrl.Result{Requeue: true} below.

controllers/loadtest_controller.go
```diff
@@ -134,7 +134,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			if updateErr := r.Status().Update(ctx, test); updateErr != nil {
 				log.Error(updateErr, "failed to update status after failure to get scenarios ConfigMap: %v", err)
 			}
-			return ctrl.Result{}, err
+			return ctrl.Result{Requeue: true}, err
```
wanlin31, line 137: Similar to above

controllers/loadtest_controller.go
```diff
@@ -305,7 +305,7 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 				if updateErr := r.Status().Update(ctx, test); updateErr != nil {
 					logWithServer.Error(updateErr, "failed to update status after failure to construct a pod for server")
 				}
-				return ctrl.Result{}, err
+				return ctrl.Result{Requeue: false}, nil
```
wanlin31, line 308: why we do not want to requeue the event due to this failure?

codeblooded: I made the distinction between two types of errors in the controller.

- There are reconciliation errors, which occur when something unexpected went wrong with the reconcile. For example, a pod could not be created. This example is something that should be retried on another reconcile.

- There are user errors, which occur when something goes wrong due to what the user has done. For example, this code uses the PodBuilder type to construct a pod for a server. It failed due to a problem with the configuration of the test itself. In this situation and in other user errors, the reconciliation should not be retried. The controller successfully "reconciled", but it updates the status to show that something has went horribly wrong. 

It is a fair to consider a failure when updating the status for user errors. In that situation, it should probably be a reconciliation error that is returned to retry. In this code, however, I just ignore the failure to update the status and log it. Perhaps it can be an enhancement, though it complicates the flow a bit.

wanlin31: Seems like this return associated to error possibly returned by PodforServer, the only error it could return is "could not determine pool for server %q (no explicit value or default", make scene not requeue this since it has nothing to do with the reconciliation process. :)

wanlin31 (approved):

PR #105, opened 2021-01-25
Fix status updates when duplicate pods are created

codeblooded:
Controllers are level-driven and operate with a cache, meaning invocations of the Reconcile method on the LoadTestReconciler may not always be able to access up-to-date information. When many invocations occur in a short timespan, the cache may not reflect the actual state of the cluster. Since a newly created pod may not be immediately noticed via the cache, multiple pod creation attempts may occur.

Previously, multiple attempts to create a pod resulted in a Kubernetes error. The test was considered to have errored. However, often the pods were created successfully and there was no mechanism to stop the test. The test could be marked as an error but continue to run successfully.

This change adds checks to prevent test errors when the controller attempts to create a duplicate pod.

### DIFF
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


wanlin31:
I think marking the LoadTest as errored when attempt to create the same pod is really our logic, actually the entire status of the LoadTest is based on our logic. Since there is nothing wrong with running the test, I think this fix should help the situation.

paulosjca (approved):

wanlin31 (approved):

PR #106, opened 2021-01-25
Feature/Ruby build and runtime images

wanlin31:
This commit adds and updates files to support Ruby tests. Build
images and ruby-example LoadTest configuration yaml file are added,
runtime image and Makefile are updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,15 @@ java-image:
 push-java-image:
 	docker push ${IMAGE_PREFIX}java:${TEST_INFRA_VERSION}
 
+# Build the Ruby build image
+ruby-build-image:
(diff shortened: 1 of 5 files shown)


wanlin31:
I verified the images use the controller running against the cluster. I am working on verify these images works for a deployed controller. Also I am a bit nervous about setting GEM_HOME manually in build_qps_worker.sh and runtime image, let me know if this is a bad practice. I will update when I get it verified with a deployed controller.

wanlin31:
Verified with deployed controller (v0.3.0).

codeblooded (approved):

codeblooded (approved):

paulosjca (approved):

wanlin31:
I have separate the part related to csharp out, this make the pr clearer. @codeblooded Thank you for pointing this out.

PR #108, opened 2021-02-02
Feature/node image

wanlin31:
I am still working on images to support Node.js tests. The build and runtime image works fine, but when I run a LoadTest, the driver logs shows that it stuck in between. Need further investigation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,6 +139,22 @@ go-image:
 push-go-image:
 	docker push ${IMAGE_PREFIX}go:${TEST_INFRA_VERSION}
 
+# Build the Node.js build image
+node-build-image:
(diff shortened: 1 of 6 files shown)


wanlin31:
The node.js support is added, however, the example node LoadTest could not be finished succeeded due to issues related to the test itself.

paulosjca (approved):

codeblooded (commented):
This LGTM, but should we merge if we do not have a way to test it is working properly? How do we know the port server issue does not live with node?

wanlin31:
Verified with deployed controller on benchmarks-dev1, reach the same status as when we run a controller agains the cluster.

wanlin31:
@codeblooded Anything wired about the checks? I failed a lot of times at go built, but I could not replicate the error locally? Any idea?

codeblooded (approved):

PR #109, opened 2021-02-03
Work around for flaky garbage collection

wanlin31:
This commit modifies the controller to not report reconcile error
when a conflict in status update has occurred. The conflict error was
not a real error since it arises from multiple reconciliations trying
to update the same object. After the change, 20 go-example loadtest
with timeoutSeconds = 30 was manually run, all passed with expected
behavior --marked as timeoutExceed at 30s. The conflict error occurred
twice, but the LoadTest finishes with expected behavior. The mechanism
of why this work around seems working and  if it really works at a large
scale was still not sure. There is no intention to make this change at
the moment, waiting for suggestions. This commit is just to keep a
record of what could possibly work.


codeblooded (approved):

paulosjca (approved):

wanlin31:
I have add some explanation of what I have done here for future reference.

PR #111, opened 2021-02-09
Add Kubernetes API client for LoadTest resources

codeblooded:
This pull request introduces an API client for interacting with LoadTests, like those found in the Kubernetes client-go library. Right now, it is only a proof-of-concept.

### DIFF
--- a/cmd/client_test/main.go
+++ b/cmd/client_test/main.go
@@ -1,0 +1,228 @@
+// Copyright 2020 gRPC authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
+	// This side-effect import is required by GKE.
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+
+	"github.com/google/uuid"
+	grpcv1 "github.com/grpc/test-infra/api/v1"
+	clientset "github.com/grpc/test-infra/clientset"
+	"github.com/grpc/test-infra/optional"
+)
+
+var (
+	schemeBuilder = runtime.NewSchemeBuilder(func(scheme *runtime.Scheme) error {
+		scheme.AddKnownTypes(grpcv1.GroupVersion,
+			&grpcv1.LoadTest{},
+			&grpcv1.LoadTestList{},
+		)
+
+		metav1.AddToGroupVersion(scheme, grpcv1.GroupVersion)
+		return nil
+	})
+	addToScheme = schemeBuilder.AddToScheme
+)
+
+func main() {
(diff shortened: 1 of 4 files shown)


codeblooded:
When ready, you should be able to perform the following actions:
```go
// given a kubernetes config, create a gRPC testing interface
grpcInterface := clientset.NewForConfig(&config)

// get a client to perform operations on version 1 Load Tests
v1Interface := grpcInterface.LoadTestV1()

// operate on tests in a specific namespace (should be default)
testGetter := v1Interface.LoadTests(corev1.NamespaceDefault)

// get a specific test by name
test, err := testGetter.Get("loadtest-name", metav1.GetOptions{})

// get all tests
testList, err := testGetter.List(metav1.ListOptions{})

// create a watcher to be notified when load tests change
watchInterface, err := testGetter.Watch(metav1.ListOptions{})
defer watchInterface.Stop()

// listen for deletion events and log the name of the test that was deleted
// see the watch godocs for details: https://pkg.go.dev/k8s.io/apimachinery/pkg/watch#Interface
eventChan := watchInterface.ResultChan()
go func() {
    event := <- eventChan
    if event.Type == watch.Deleted {
        test := event.(*grpcv1.LoadTest)
        fmt.Printf("deleted test named %q", test.Name)   
    }
}()
```

codeblooded:
I will add a Create method to the aforementioned LoadTestGetter interface, so you will be able to also run:

```go
// create a test given a valid grpcv1.LoadTest named "test"
createdTest, err := testGetter.Create(&test, metav1.CreateOptions{})
```

Let me know if we need `Update`, `UpdateStatus` and `Delete`/`DeleteCollection`. I do  not think we should support `Patch`.

codeblooded:
I added `Create` and `Delete`:

```go
// create a test given a valid grpcv1.LoadTest named "test"
createdTest, err := testGetter.Create(&test, metav1.CreateOptions{})

// delete a test given the name of a valid test
err = testGetter.Delete(test.Name, metav1.DeleteOptions{})
```

codeblooded:
All client methods are working except for the Watcher, which is producing the following error:

```
2021/02/11 15:09:50 event: {ERROR &Status{ListMeta:ListMeta{SelfLink:,ResourceVersion:,Continue:,RemainingItemCount:nil,},Status:Failure,Message:an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: no kind \"LoadTest\" is registered for the internal version of group \"e2etest.grpc.io\" in scheme \"pkg/runtime/scheme.go:101\"") has prevented the request from succeeding,Reason:InternalError,Details:&StatusDetails{Name:,Group:,Kind:,Causes:[]StatusCause{StatusCause{Type:UnexpectedServerResponse,Message:unable to decode an event from the watch stream: unable to decode watch event: no kind "LoadTest" is registered for the internal version of group "e2etest.grpc.io" in scheme "pkg/runtime/scheme.go:101",Field:,},StatusCause{Type:ClientWatchDecoding,Message:unable to decode an event from the watch stream: unable to decode watch event: no kind "LoadTest" is registered for the internal version of group "e2etest.grpc.io" in scheme "pkg/runtime/scheme.go:101",Field:,},},RetryAfterSeconds:0,UID:,},Code:500,}}
panic: interface conversion: runtime.Object is *v1.Status, not *v1.LoadTest
```

Investigating and will mail for review once fixed.

codeblooded:
Since Watch events are not working, I amended to support only Create, Get, List and Delete. When the watch issue is resolved, we can open a follow-up PR to add its functionality.

This change can be tested locally (with access to a real Kubernetes cluster) by running `go run cmd/client_test/main.go`. This script tests the above functionality. In addition, it serves as an example of how to use the client.

cmd/client_test/main.go
```diff
@@ -0,0 +1,228 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
+	// This side-effect import is required by GKE.
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
codeblooded, line 32: Note: this script is coupled to GKE and will not work in other Kubernetes clusters. In order to use EKS, AKS, etc... you would need to replace "gcp" with the appropriate plugin.

paulosjca (approved):

PR #112, opened 2021-02-11
Feature/PHP build and runtime images

wanlin31:
This commit adds dockerfils and script to support PHP tests. Build
images and PHP-example LoadTest configuration yaml file are added,
runtime image and Makefile are updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,11 +163,27 @@ java-image:
 # Push the Java runtime image to a docker registry
 push-java-image:
 	docker push ${IMAGE_PREFIX}java:${TEST_INFRA_VERSION}
-
 # Build the Ruby build image
 ruby-build-image:
 	docker build -t ${BUILD_IMAGE_PREFIX}ruby:${TEST_INFRA_VERSION} \
 		containers/init/build/ruby
+
+# Push the PHP runtime image to a docker registry
+push-php-build-image:
+	docker push ${BUILD_IMAGE_PREFIX}php:${TEST_INFRA_VERSION}
+
+# Build the PHP runtime image
+php-image:
+	docker build -t ${IMAGE_PREFIX}php:${TEST_INFRA_VERSION} \
+		containers/runtime/php
+
+# Push the PHP runtime image to a docker registry
+push-php-image:
+	docker push ${IMAGE_PREFIX}php:${TEST_INFRA_VERSION}
+# Build the PHP build image
+php-build-image:
+	docker build -t ${BUILD_IMAGE_PREFIX}php:${TEST_INFRA_VERSION} \
+		containers/init/build/php
 
 # Push the Ruby runtime image to a docker registry
 push-ruby-build-image:
@@ -210,6 +226,8 @@ all-images: \
 	node-build-image \
 	node-image \
 	python-image \
+	php-build-image \
+	php-image \
 	ruby-build-image \
 	ruby-image \
 	csharp-build-image \
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,7 @@ php-image:
 # Push the PHP runtime image to a docker registry
 push-php-image:
 	docker push ${IMAGE_PREFIX}php:${TEST_INFRA_VERSION}
+
 # Build the PHP build image
 php-build-image:
 	docker build -t ${BUILD_IMAGE_PREFIX}php:${TEST_INFRA_VERSION} \
--- a/Makefile
+++ b/Makefile
@@ -163,28 +163,11 @@ java-image:
 # Push the Java runtime image to a docker registry
 push-java-image:
 	docker push ${IMAGE_PREFIX}java:${TEST_INFRA_VERSION}
+
 # Build the Ruby build image
 ruby-build-image:
 	docker build -t ${BUILD_IMAGE_PREFIX}ruby:${TEST_INFRA_VERSION} \
 		containers/init/build/ruby
-
-# Push the PHP runtime image to a docker registry
-push-php-build-image:
-	docker push ${BUILD_IMAGE_PREFIX}php:${TEST_INFRA_VERSION}
-
-# Build the PHP runtime image
-php-image:
-	docker build -t ${IMAGE_PREFIX}php:${TEST_INFRA_VERSION} \
-		containers/runtime/php
-
-# Push the PHP runtime image to a docker registry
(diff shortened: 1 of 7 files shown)


wanlin31:
Still working on this. The build step works, but some executables or dependencies have be installed in a not shared volume so that when the build container dies the run container has no access to those files.  
Next step: Figure out how to make built worker and its dependencies all in the shared volume.

wanlin31:
The workers run okay, however, seems like the driver is not able to connect to the workers through its port. Probably would like to discuss the next step in our meeting.

wanlin31:
@codeblooded Could you help me to take a look if anything I missed? I got the workers run okay, however the driver was not able to connect to the workers. Thank you. :)

config/samples/php_example_loadtest.yaml
```diff
@@ -0,0 +1,131 @@
...
+
+  # Users can specify multiple clients. They are bound by the
+  # number of nodes.
+  clients:
+  - language: php
+    clone:
+      repo: https://github.com/grpc/grpc.git
+      gitRef: master
+    build:
+      image: gcr.io/grpc-testing/wanlin/init/build/php
```
wanlin31, line 43: Need to clean this up later.

Makefile
```diff
@@ -147,12 +147,28 @@ java-image:
...
+
+# Build the PHP runtime image
+php-image:
+	docker build -t ${IMAGE_PREFIX}php:${TEST_INFRA_VERSION} \
+		containers/runtime/php
+
+# Push the PHP runtime image to a docker registry
+push-php-image:
+	docker push ${IMAGE_PREFIX}php:${TEST_INFRA_VERSION}
+# Build the PHP build image
```
codeblooded, line 167: nit: missing a space between lines 166 and 167. In addition, let's keep push- and the build of the images together. So, the target push-ruby-build-image will follow ruby-build-image and push-php-image will follow php-image and so on.

containers/init/build/php/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+FROM php:7.2.22-buster
+RUN mkdir -p /src/workspace
+WORKDIR /src/workspace
+
+RUN apt-get update && apt-get install -y \
+  git \
+  zlib1g-dev \
+  build-essential \
+  lcov \
+  rubygems \
```
codeblooded, line 24: In my understanding, rubygems should be installed alongside ruby when the installation works properly. This is the case for rbenv at least. I believe this was also the behavior for rvm, but it has been years since I have used it.

codeblooded (changes requested):
A few small questions, but LGTM.

Makefile
wanlin31: Right, thanks for pointing this out!

containers/init/build/php/Dockerfile
wanlin31: I have tried without installing rubygems, it just doesn't work...

containers/init/build/php/build_qps_worker.sh
```diff
@@ -0,0 +1,42 @@
...
+
+cd /src/workspace/src/php/tests/qps
+
+composer install
+
+cd ../../../../third_party/protobuf/php/ext/google/protobuf
+
+phpize
+./configure
+
```
wanlin31, line 36: @codeblooded Actually I am pretty unsure about his part, I have took the content of their script and put them here, however I am not so sure if any off the lines are redundant, do you have some idea of that?

containers/init/build/php/Dockerfile
wanlin31: I think I have tried without it and doesn't work, but I think you could be right, I have tried too many combinations and lost track of what I have altered, let me try again and fix it.

wanlin31: You are right, we do not need rubygems installed , fixed

containers/init/build/php/build_qps_worker.sh
codeblooded: I do not think it is redundant b/c we change directory in line 32. So, they configure grpc in lines 25 and 26 and protobuf in lines 34 and 35. I don't know why they would do grpc before protobuf, but the redundancy isn't wrong.

codeblooded:
@wanlin31 I'm not sure why the driver is not connecting. Do you have any error logs you can share with me or container statuses?

wanlin31:
Yes, @codeblooded, thank you for helping me take a look of this.

wanlin31:
Verified with a deployed controller on benchmarks cluster, workers are up and running at 14mins, but driver could not connect to the workers. Same as when we run controller against the cluster.

Makefile
codeblooded: nit: I'm not sure if this is resolved. I think images and their push should be adjacent. So, for example, the `push-php-build-image` target is defined in line 156. The image that it is pushing, `php-build-image` is defined in line 169. I'd like to keep them together, in an order, like-so:

```
php-build-image:
push-php-build-image:
php-image:
push-php-image:
```

This ordering makes the file a little easier to read and comprehend.

containers/init/build/php/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+RUN /bin/bash -l -c "echo 'gem: --no-document' > ~/.gemrc"
+RUN /bin/bash -l -c "echo 'export PATH=/usr/local/rvm/bin:$PATH' >> ~/.bashrc"
+RUN /bin/bash -l -c "echo 'rvm --default use ruby-2.5' >> ~/.bashrc"
+RUN /bin/bash -l -c "gem install bundler --no-document -v 1.9"
+
+# Install composer
+RUN curl -sS https://getcomposer.org/installer | php
+RUN mv composer.phar /usr/local/bin/composer
+
+
```
codeblooded, line 45: nit: extra whitespace... let's keep it a 1 line break, not 2

codeblooded (commented):
I have two nits, but otherwise LGTM.

containers/init/build/php/Dockerfile
wanlin31: Right, I guess I might have it accidentally resolved, would definitely fix them. Thank you :)

codeblooded (approved):

paulosjca (approved):

PR #114, opened 2021-02-19
Remove unnecessary call to t.Fatalf from prototype test code.

paulosjca:
This change removes an unnecessary call to `t.Fatalf` from prototype test code. The call happens when an event is generated in the test that does not correspond to an existing component. Instead of calling `t.Fatalf`, the test sets the component name to `_UNKNOWN_`. This will cause an error later in the test, when the tests checks if the events match the test expectations.

This change is needed to prevent an error flagged by `go vet`:

```
go vet ./...
benchmarks/svc/orch/executor_test.go:141:7: call to (*T).Fatalf from a non-test goroutine
make: *** [Makefile:75: vet] Error 2
```


codeblooded (approved):
LGTM. CI failed due to integration test environment flakiness. Restarting your CI run.

wanlin31 (approved):

PR #116, opened 2021-02-25
Modify cxx image to start the start_port_server.py first

wanlin31:
This commit adds a script to start start_port_server.py first,
this step is to allocate ports used during the test.


wanlin31:
This commits add scripts tp start start_server_port.py, not sure if this is what we want to pursue.

wanlin31:
This is no longer needed, we directly feed the port to the servers.

PR #119, opened 2021-03-19
Add server port flag and explanation to example

codeblooded:
The C++ workers in grpc/grpc rely on a server to find available ports when no explicit port is given. Unfortunately, this script relies on Python's subprocess package. The subprocess package has some compatibility issues with Docker, creating an issue where the port server could not start.

To temporarily resolve this problem, this commit explicitly introduces a --server_port flag in the run section of the server for the C++ example LoadTest YAML file. Specifying this flag disables the need for a port server, bypassing the problematic code.

### DIFF
--- a/config/samples/cxx_example_loadtest.yaml
+++ b/config/samples/cxx_example_loadtest.yaml
@@ -30,6 +30,10 @@ spec:
         args: ["build", "//test/cpp/qps:qps_worker"]
       run:
         command: ["bazel-bin/test/cpp/qps/qps_worker"]
+        # For C++, we must specify a --server_port flag for now. This
+        # requirement exists, because there are problems with the
+        # port server in grpc/grpc running on Docker.
+        args: ["--server_port=10010"]
 
   # Users can specify multiple clients. They are bound by the
   # number of nodes.


paulosjca (approved):

wanlin31 (approved):
I also have tried this with pre-built cxx workers, worked just great!

wanlin31:
Do you want to add a release tag on this one?

codeblooded:
> Do you want to add a release tag on this one?

I'm not sure. It solves the problem, but it's just an example configuration change. It's more of a workaround. I'll merge and let @paulosjca add the release notes label if he feels it's appropriate. I'm indifferent.

jtattermusch:
@codeblooded  @wanlin31 
I haven't looked into the C++ port server issue in detail, but this tip might come handy:
- the grpc build allows one to build with the `GRPC_PORT_ISOLATED_RUNTIME` define.
https://github.com/grpc/grpc/blob/86f93801e5d452cff40d6958d21a8eb211da0654/test/core/util/port_isolated_runtime_environment.cc#L29
The semantics is basically "we know we're gonna be running in environment where port collision is not going to happen (e.g. under docker container), so instead of using port server, simply pick a port randomly".
- if running port_server under docker is a problem (I wouldn't be surprised), we can build C++ qps_worker with GRPC_PORT_ISOLATED_RUNTIME=1 and that will eliminate the need for specifying the `--server_port` arg?
- if we decide to use this workaround (leaving up to you), there should a be a comment somewhere that explains what's being done and why.

We use this for some of our tests that we know are running in an isolated environment:
https://github.com/grpc/grpc/blob/03adef1d3cb6a796f9a910794dfbcea48f510ba8/tools/remote_build/rbe_common.bazelrc#L43

PR #122, opened 2021-04-05
Add prebuilt image for nodejs

ybbbby:
Following the pattern of go and cxx, add prebuilt image and sample loadtest for nodejs.
The image size:
```
6d6d9eee67d4   7 days ago    /bin/sh -c #(nop) COPY file:a70d9190681e6dc4…   52B       
684ddb59c632   7 days ago    /bin/sh -c #(nop) COPY file:a9e20847a5130dd6…   684B      
4bb802387063   7 days ago    /bin/sh -c #(nop) COPY file:d12475199ea2307c…   114MB 
```
So approximately 114MB would be pushed to gcr.io.
Using prebuilt image, we can save around 2m30s for performance test(skipping clone and build stage).


Makefile
```diff
@@ -151,7 +151,7 @@ push-node-build-image:
 node-image:
 	docker build -t ${IMAGE_PREFIX}node:${TEST_INFRA_VERSION} containers/runtime/node
 
-# Push the Go runtime image to a docker registry
+# Push the Node.js runtime image to a docker registry
```
wanlin31, line 154: Good catch!

config/samples/node_example_loadtest_with_pre_built_workers.yaml
```diff
@@ -0,0 +1,127 @@
...
+  # initial version only supports 1 server to limit scope. Servers
+  # is an array for future expansion.
+  #
+  # There are many designs and systems to pursue load balancing,
+  # organizing and monitoring a mesh of servers. Therefore, this
+  # will likely be expanded in the future.
+  servers:
+  - language: node
+    run:
+      image: gcr.io/grpc-testing/e2etest/prebuild/node
```
wanlin31, line 26: I usually test these images in my repo, like gcr.io/grpc-testing/wanlin/prebuild/node instead of the official repo for release. I think @codeblooded know more if this is okay.

config/samples/node_example_loadtest_with_pre_built_workers.yaml
```diff
@@ -0,0 +1,127 @@
...
+      command: [ "node" ]
+      args: [ "-r", "/execute/test/fixtures/native_native.js", "/execute/test/performance/worker.js", "--benchmark_impl=grpc"]
+
+  # Users can specify multiple clients. They are bound by the
+  # number of nodes.
+  clients:
+  - language: node
+    run:
+      image: gcr.io/grpc-testing/e2etest/prebuild/node
```
wanlin31, line 35: Same as previous comment.

config/samples/node_example_loadtest_with_pre_built_workers.yaml
ybbbby: agree, done

config/samples/node_example_loadtest_with_pre_built_workers.yaml
ybbbby: done

containers/pre_built_workers/node/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+COPY --from=0 /pre/test/fixtures/native_native.js /execute/test/fixtures/native_native.js
+ENV NODE_OPTIONS='--require /execute/test/fixtures/native_native.js'
+RUN apt-get update && apt-get install -y \
+  bash \
+  curl \
+  git \
+  time && \
+  apt-get clean
+
+CMD ["bash"]
```
wanlin31, line 51: nit: I think we should have an empty line at the end of the file.

wanlin31: My bad, I did not follow the proper form when my PR was in a draft state :)

containers/pre_built_workers/node/build_qps_worker.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+npm install -production
```
codeblooded, line 18: nit: Let's use 2 dashes for long-form flags and 1 dash for short-form flags. For example, `-v` and `--verbose`.

```suggestion
npm install --production
```

containers/pre_built_workers/node/pkg_config.json
```diff
@@ -0,0 +1 @@
+{"assets" : "test/**/*"}
```
codeblooded, line 1: nit: Let's end every file UNIX style with an empty newline like @wanlin31 suggested.

codeblooded (commented):
Two comments, but you're doing great @ybbbby. And, I think @wanlin31 is doing an incredible job as a code reviewer and unofficial mentor 😄

containers/pre_built_workers/node/Dockerfile
ybbbby: Thanks for reviewing! Done.

containers/pre_built_workers/node/build_qps_worker.sh
ybbbby: Thanks for reviewing:) Done.

containers/pre_built_workers/node/pkg_config.json
ybbbby: Done.

containers/pre_built_workers/node/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM node:10-buster
+
+RUN mkdir -p /pre
+WORKDIR /pre
+
+ARG REPOSITORY=grpc/grpc-node
```
wanlin31, line 20: As we have talked about since it is a ARG, it takes a bit work to access after the build time. We might want to save them somewhere for future access.  I had something like 
```
RUN echo 'COMMIT SHA' > GRPC_GIT_COMMIT.txt
RUN git rev-parse $GITREF >> GRPC_GIT_COMMIT.txt
```
in pr #113. You could do something else too. :) Also it might be helpful to read the comments from Jan for pr#113. :)

containers/pre_built_workers/node/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# Copy node modules to a new image
+FROM debian:buster
+
+RUN mkdir -p /pre
+WORKDIR /execute
+COPY --from=0 /pre/worker-linux /execute
+COPY --from=0 /pre/test/fixtures/native_native.js /execute/test/fixtures/native_native.js
+ENV NODE_OPTIONS='--require /execute/test/fixtures/native_native.js'
+RUN apt-get update && apt-get install -y \
```
wanlin31, line 44: I have checked the size of the layers, line 44 is about 125MB.
nit: 
(1) Since we have discussed that if any underlying layer changes, the subsequent layer would be update and saved again, taking extra space. It would be more space efficient if line 44 happens before line 41. 
(2) Do we need all of them?  I know these probably came from the original node.js runtime Dockerfile,  since those runtime images only need to push once per release, the optimization requirement is not as strict as pre-built images.  We care about this size even we might only save once of this layer in gcr.io, but we have to push and pull these images at the test time, the bigger the size is the longer it takes to push to gcr.io and download to a pod.

Let me know if this makes scene.

containers/pre_built_workers/node/build_qps_worker.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+npm install --production
```
wanlin31, line 18: Good finding! I inherit what gRPC had (https://github.com/grpc/grpc/blob/master/tools/run_tests/performance/build_performance_node.sh) but I think have a --production flag to prevent installation of devDependencies is good.

containers/pre_built_workers/node/Dockerfile
ybbbby: Done! thanks for reference, those bazel-related comments help me improving the prebuilt image for python!

containers/pre_built_workers/node/Dockerfile
ybbbby: Done! Discussed offline. We decided to get rid of the whole apt-get update b/c the binary does not have dependencies on it.

wanlin31:
The test is flaky at the moment it fails sometimes, when it fails you can just click `Show all checks`->`Details`->it will be directed to another page, click the `Re-run jobs` -> select `Re-run all jobs`, you will then be triggering another round of the test.

ybbbby:
> The test is flaky at the moment it fails sometimes, when it fails you can just click `Show all checks`->`Details`->it will be directed to another page, click the `Re-run jobs` (it is on the top right corner, just above the code block ) -> select `Re-run all jobs`, you will then be triggering another round of the test.

Thanks! Now the 'Re-run jobs' button shows up after me getting authorized :)

config/samples/node_example_loadtest_with_pre_built_workers.yaml
```diff
@@ -0,0 +1,124 @@
...
+  # initial version only supports 1 server to limit scope. Servers
+  # is an array for future expansion.
+  #
+  # There are many designs and systems to pursue load balancing,
+  # organizing and monitoring a mesh of servers. Therefore, this
+  # will likely be expanded in the future.
+  servers:
+  - language: node
+    run:
+      image: gcr.io/grpc-testing/donghao/prebuilt/node
```
wanlin31, line 26: I think we could leave it as 'image: <current_image_repository>' or something else,  since we will use tools/run_tests/performance/loadtest_template.py to alter the field into a variable that could be changed in https://github.com/grpc/grpc/blob/master/tools/run_tests/performance/loadtest_config.py. The examples employing prebuilt images could be served as template to generate the basic template. Let me know if you think this makes scene. :) @codeblooded, do you think this make scene?

wanlin31 (commented):
LGTM. :)

The pr looks good to me, I tried to run it with benchmark-prod cluster, get to the same stage as our regular images. Once you get the size for each layer print out this PR should be good.

You should wait for @paulosjca and @codeblooded  to take a look, I might have overlooked something.

wanlin31 (approved):

paulosjca (approved):

config/samples/node_example_loadtest_with_pre_built_workers.yaml
paulosjca: This has been addressed: configuration is now a template, and the variables in it will be passed unchanged to the mutli-language template.

PR #125, opened 2021-04-09
Update READY_TIMEOUT_ARG to be a environmental variable

wanlin31:
This commit adds READY_TIMEOUT_ARG as environmemntal variable in
Makefile, makes the deployment process easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ CLEAN_IMG ?= ${IMAGE_PREFIX}cleanup:${TEST_INFRA_VERSION}
 #${IMAGE_PREFIX}cleanup_agent:${TEST_INFRA_VERSION}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
+# Set timeout for ready container
+READY_TIMEOUT_ARG ?= "20m"
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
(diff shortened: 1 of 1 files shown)


wanlin31:
I will verify this works before open for review.

wanlin31:
Verified by building and pushing a image successfully.

Makefile
```diff
@@ -18,6 +18,8 @@ CLEAN_IMG ?= ${IMAGE_PREFIX}cleanup:${TEST_INFRA_VERSION}
 #${IMAGE_PREFIX}cleanup_agent:${TEST_INFRA_VERSION}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
+# Set timeout for ready container
+READY_TIMEOUT_ARG ?=20m
```
codeblooded, line 22: nit: Add a space between equal and quote to escape any system that may interpret it as an integer.

```suggestion
READY_TIMEOUT_ARG ?= "20m"
```

codeblooded (approved):
Looks good to me. I just have one style nit, but otherwise I think it's a good enhancement.

Makefile
wanlin31: Thank you for pointing that out!

wanlin31:
This is should have been merged at the time of release. Lesson learned: once the release is announced, nothing should be included into that release, anything later merged should be included in the next release.

PR #127, opened 2021-04-09
Add prebuilt image for python

ybbbby:
Python image is built by Bazel from grpc/grpc, and can be further optimized.
After built, bone structure/binary is extracted to a new image to shrink the final image's size.
The build procedure mimic the init-containers' clone, build and run steps.
The image size:
```
30b5098c98b6   30 seconds ago   /bin/sh -c #(nop) COPY file:1288b1d07ba39958…   52B       
2cc577ed0a38   22 hours ago     /bin/sh -c #(nop) COPY file:2119411c1149daf8…   4.99MB  
```
So approximately 8.43MB would be pushed to gcr.io.
Using prebuilt image, we can complete the performance test within 50s while we need 7m32s to finish using multi-stage images. Thus, we can save approximately 6m40s.


wanlin31:
I think for this great work you have done, it will be nice to let reviewers know that: 
(usually what I did, hopefully helpful :) )

(1) You have verified with benchmark-prod (deployed controller) or benchmark-dev1 (deployed controller or local controller run against cluster) using this image, and reach to the same stage as regular images. (I usually have this statement in the comment for the previous images, it is up to you how you want to verify it, but at least people know you have verified it)
(2) What is the size  for the binary we need to push each time to the repositories.
(3) What is the time we saved for using pre-built image.

I have not went through cxx for these step in the comments is because it is part of the investigation and have been verified and reported several times, and for go, I have not yet mark the pr as ready, I will probably do the same thing as I described.

wanlin31:
This apply to all the images, I leave it here so I won't flood all your PRs. :)

ybbbby:
> I think for this great work you have done, it will be nice to let reviewers know that:
> (usually what I did, hopefully helpful :) )
> 
> (1) You have verified with benchmark-prod (deployed controller) or benchmark-dev1 (deployed controller or local controller run against cluster) using this image, and reach to the same stage as regular images. (I usually have this statement in the comment for the previous images, it is up to you how you want to verify it, but at least people know you have verified it)
> (2) What is the size for the binary we need to push each time to the repositories.
> I have done the analysis of this one, we will push ~8.43MB each time to gcr.io, looks great.
> 
> ```
> 64227fbc6206   10 minutes ago   /bin/sh -c #(nop) COPY file:838ec40eee1c3cf3…   52B       
> ed81bd397642   10 minutes ago   /bin/sh -c #(nop) COPY file:016ebc82536415ae…   8.43MB 
> ```
> 
> (3) How much is the time we saved for using pre-built image.
> 
> I have not went through cxx for these step in the comments is because it is part of the investigation and have been verified and reported several times, and for go, I have not yet mark the pr as ready, I will probably do the same thing as I described.

Thanks for pointing out!

paulosjca (approved):

containers/pre_built_workers/python/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN git clone https://github.com/$REPOSITORY.git .
+RUN git submodule update --init
+RUN git checkout $GITREF
+
+# Save commit sha for debug use
+RUN echo 'COMMIT SHA' > GRPC_GIT_COMMIT.txt
+RUN git rev-parse $GITREF >> GRPC_GIT_COMMIT.txt
+
+RUN mkdir /executables
+RUN bazel --output_user_root=/executables build --nobuild_runfile_links --nobuild_runfile_manifests --build_python_zip //src/python/grpcio_tests/tests/qps:qps_worker
```
wanlin31, line 32: As we have spoke in the meeting, you might want to have -c opt option. See reference: https://github.com/grpc/grpc/blob/4768210ec860fd9a5ecba14cf11de147c32873fd/tools/run_tests/performance/build_performance.sh#L72-L77

wanlin31: Other than that LGTM!

ybbbby: Done! Thanks for reviewing:DD

config/samples/python_asyncio_example_loadtest.yaml
```diff
@@ -0,0 +1,134 @@
...
+  # cluster. There are ways we can circumvent naming clashes, such
+  # as using namespaces or dynamically assigning names.
+  name: python-asyncio-example
+
+  # As a custom resource, it behaves like a native kubernetes object.
+  # This means that users can perform CRUD operations through the
+  # Kubernetes API or kubectl. In addition, it means that the user
+  # can set any metadata on it.
+  labels:
+    language: python
```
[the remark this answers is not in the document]
wanlin31: I think we also need to update the default template (https://github.com/grpc/test-infra/blob/master/config/defaults_template.yaml) to make sure it knows where to find the images.
something like 
```
 - language: python_asyncio
```

config/samples/python_asyncio_example_loadtest.yaml
```diff
@@ -0,0 +1,134 @@
+apiVersion: e2etest.grpc.io/v1
```
wanlin31, line 1: I think it is okay this time, you might want to avoid mixing things up in a single PR in the future. For example, this PR is dedicated for adding prebuilt images for python, it made more or less scene to also include prebuilt images for python_asyncio since they are both included by word python. But for adding python_asyncio_example_loadtest.yaml it makes more scene to have its own PR. This way, when we select PRs to do a release,  by reading the name of PR we have a basic understanding of what it is, also it gives us options to include the change this time or next. :)

codeblooded (approved):

wanlin31:
I think Paulo already mentioned things I want to talk about. LGTM.

wanlin31 (approved):

config/samples/python_asyncio_example_loadtest.yaml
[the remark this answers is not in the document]
wanlin31: Only thing wanted to mention, for regular 3-stage building, there is no change of the image, just the command field in build session in the example loadtest configuration is different.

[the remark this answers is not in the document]
ybbbby: Thanks for pointing out! I have moved all python-asyncio related change to another pr, see: https://github.com/grpc/test-infra/pull/142

config/samples/python_asyncio_example_loadtest.yaml
ybbbby: Thanks, this definitely make sense. Done.

wanlin31 (approved):

PR #129, opened 2021-04-09
Add prebuilt image for ruby

ybbbby:
Ruby image is built by bundle, rake, cmake from grpc/grpc.
After built, bone structure/binary is extracted to a new image to shrink the final image's size.
The build procedure mimic the init-containers' clone, build and run steps.
The image size:
```
55dde96d8841   7 days ago    /bin/sh -c #(nop) COPY file:91288581a205eb37…   52B       
03cc5e005684   7 days ago    /bin/sh -c #(nop) COPY dir:17f0ec771f5339353…   264kB     
7bfaeed7ea16   7 days ago    /bin/sh -c #(nop) COPY dir:7faa2ab1498217156…   20.8MB    
6d6fc038cb97   10 days ago   /bin/sh -c #(nop) COPY dir:9afd81795b8e120b8…   26.7MB  
```
So approximately 47.5MB would be pushed to gcr.io.
Using prebuilt image, we can complete the performance test within 45s while we need 7m to finish using multi-stage images. Thus, we can save approximately 6m15s.

### DIFF
--- a/config/samples/ruby_example_loadtest_with_pre_built_workers.yaml
+++ b/config/samples/ruby_example_loadtest_with_pre_built_workers.yaml
@@ -23,7 +23,7 @@ spec:
   servers:
   - language: ruby
     run:
-      image: gcr.io/grpc-testing/donghao/prebuilt/ruby
+      image: ${prebuilt_image_prefix}/ruby
       command: [ "ruby" ]
       args: [ "/execute/src/ruby/qps/worker.rb" ]
 
@@ -32,7 +32,7 @@ spec:
   clients:
   - language: ruby
     run:
-      image: gcr.io/grpc-testing/donghao/prebuilt/ruby
+      image: ${prebuilt_image_prefix}/ruby
       command: [ "ruby" ]
       args: [ "/execute/src/ruby/qps/worker.rb" ]
 
--- a/config/samples/ruby_example_loadtest_with_pre_built_workers.yaml
+++ b/config/samples/ruby_example_loadtest_with_pre_built_workers.yaml
@@ -1,0 +1,121 @@
+apiVersion: e2etest.grpc.io/v1
+kind: LoadTest
+metadata:
+  # Every load test instance must be assigned a unique name on the
+  # cluster. There are ways we can circumvent naming clashes, such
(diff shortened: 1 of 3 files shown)


config/samples/ruby_example_loadtest_with_pre_built_workers.yaml
```diff
@@ -0,0 +1,121 @@
...
+  # initial version only supports 1 server to limit scope. Servers
+  # is an array for future expansion.
+  #
+  # There are many designs and systems to pursue load balancing,
+  # organizing and monitoring a mesh of servers. Therefore, this
+  # will likely be expanded in the future.
+  servers:
+  - language: ruby
+    run:
+      image: gcr.io/grpc-testing/donghao/prebuilt/ruby
```
codeblooded, line 26: We don't want to mention the explicit cloud project 'grpc-testing' and folder 'donghao' here. This sample will not be runnable by anyone who does not have GCR read access here. This should be made into a fake image path, or it should be used to reopen discussions about making our container images public. /cc @grpc/test-infra

codeblooded (changes requested):

config/samples/ruby_example_loadtest_with_pre_built_workers.yaml
paulosjca: I would change these files into templates. In this case, the location could be ${prebuilt_image_prefix} of some such. I would use Python template notation, since these templates will be processed by loadtest_template.py in grpc/grpc (compare that with the templates in https://github.com/grpc/test-infra/blob/master/config/defaults_template.yaml, which are processed by a go script from our makefiles.

If we make these files into templates, I would create a new folder config/samples/templates for them, and perhaps add a README.md there explaining that these are templates and should be run through loadtest_template.py first.

I would ultimately run all languages through loadtest_template.py and create a new master template in grpc/grpc, similar to the one in https://github.com/grpc/grpc/pull/25976.

ybbbby: done.

paulosjca (approved):

wanlin31 (approved):

codeblooded (approved):

PR #130, opened 2021-04-12
Feature/Add pre-built go example and Dockerfile

wanlin31:
This commits add the Dockerfile that contains a pre-built
go-worker and a example load test configuration file example to
use such pre-built workers.

### DIFF
--- a/config/samples/go_example_loadtest_with_pre_built_workers.yaml
+++ b/config/samples/go_example_loadtest_with_pre_built_workers.yaml
@@ -1,0 +1,128 @@
+apiVersion: e2etest.grpc.io/v1
+kind: LoadTest
+metadata:
+  # Every load test instance must be assigned a unique name on the
+  # cluster. There are ways we can circumvent naming clashes, such
+  # as using namespaces or dynamically assigning names.
+  name: prebuilt-go-example
+
+  # As a custom resource, it behaves like a native kubernetes object.
+  # This means that users can perform CRUD operations through the
+  # Kubernetes API or kubectl. In addition, it means that the user
+  # can set any metadata on it.
+  labels:
+    language: go
+spec:
+  # The user can specify servers to use when running tests. The
+  # initial version only supports 1 server to limit scope. Servers
+  # is an array for future expansion.
+  #
+  # There are many designs and systems to pursue load balancing,
+  # organizing and monitoring a mesh of servers. Therefore, this
+  # will likely be expanded in the future.
+  servers:
+    - language: go
+      run:
+        image: ${prebuilt_image_prefix}/go:${prebuilt_image_tag}
+        command: ["/executable/worker"]
+
+
+  # Users can specify multiple clients. They are bound by the
+  # number of nodes.
+  clients:
+    - language: go
+      run:
+        image: ${prebuilt_image_prefix}/go:${prebuilt_image_tag}gcloud container clusters get-credentials benchmarks-dev1 \
+          --zone us-central1-b \
+          --project grpc-testing
+        command: ["/executable/worker"]
+
+  # We can optionally specify where to place the results. The
+  # controller will attempt to mount a service account in the driver.
+  # This can be used for uploading results to GCS or BigQuery.
+  # results:
+  #   bigQueryTable: "example-project.foo.demo_dataset"
+
+  # timeoutSeconds is an integer field that indicates the longest time a test
+  # is allowed to run, in seconds. Tests that run longer than the given value
+  # will be marked as Errored and will no longer be allocated resources to run.
+  # For example: timeoutSeconds: 900 indicates the timeout of this test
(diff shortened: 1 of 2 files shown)


wanlin31:
Size analysis for pre-built go image, the actual size to push to container registry is ~16.5MB locally.
```
IMAGE          CREATED          CREATED BY                                      SIZE      COMMENT
7b1f4a53a859   53 seconds ago   /bin/sh -c #(nop)  CMD ["bash"]                 0B        
d1d771fbfcf2   54 seconds ago   /bin/sh -c #(nop) COPY dir:14198f9003a41c5a4…   57kB      
60a8ac55dea4   54 seconds ago   /bin/sh -c #(nop) COPY file:c2dff35212697eaa…   16.1MB    
792ded2aca1b   55 seconds ago   /bin/sh -c #(nop) WORKDIR /executable           0B        
ced88af36f9b   55 seconds ago   /bin/sh -c mkdir -p /executable                 0B        
0d587dfbc4f4   2 weeks ago      /bin/sh -c #(nop)  CMD ["bash"]                 0B        
<missing>      2 weeks ago      /bin/sh -c #(nop) ADD file:e18bc3e10e7c743f1…   114MB     
```

wanlin31:
Verified on benchmarks-prod.

config/samples/go_example_loadtest_with_pre_built_workers.yaml
```diff
@@ -0,0 +1,128 @@
...
+
+
+  # Users can specify multiple clients. They are bound by the
+  # number of nodes.
+  clients:
+    - language: go
+      run:
+        image: ${prebuilt_image_prefix}/go:${prebuilt_image_tag}gcloud container clusters get-credentials benchmarks-dev1 \
+          --zone us-central1-b \
+          --project grpc-testing
```
codeblooded, line 37: It looks like some of your history got included. I think you meant this:

```suggestion
        image: ${prebuilt_image_prefix}/go:${prebuilt_image_tag}
```

containers/pre_built_workers/go/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Build go excutables
+RUN go build -o /executable/bin/worker ./benchmark/worker
+
+# Copy go excuatable to a new image
+FROM debian:buster
+
+RUN mkdir -p /executable
+WORKDIR /executable
+COPY --from=0 /executable/bin/worker /executable
```
codeblooded, line 39: nit: I think renaming the file from `/executable/bin/worker` in one container to `/executable/worker` in another is a tad confusing. It would be nice if we could make this consistent. They both could be `/executable/worker`. 

I assume we are using the `/executable` path to not cause conflict with `/src/workspace`? It seems a little confusing and inconsistent b/c we never provide an `/executable` mounts.

codeblooded (changes requested):

containers/pre_built_workers/go/Dockerfile
codeblooded: /src/workspace does not necessarily have to be code. It was meant to be the "workspace," the current working directory for whatever we must test. Unfortunately, bazel broke this some by symlinking.

config/samples/go_example_loadtest_with_pre_built_workers.yaml
wanlin31: Right! Thank you for pointing this out!

containers/pre_built_workers/go/Dockerfile
wanlin31: Right, I change the path name due to the conflict. Since /src/workspace shares the volume inherited from previous stages, if we use `/src/workspace`, regardless what we have saved in the image it will became empty (inherited from clone - build- step, which in the pre-built case, we do not have anything there). 

During the process of working on pre-built image, I found out that if we copy those symlinks, the actual binaries would be copied. If really need we could do that optimization later. :) 

Thank you for helping me with the formats!

paulosjca (approved):

containers/pre_built_workers/go/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN go build -o /executable/bin/worker ./benchmark/worker
+
+# Copy go excuatable to a new image
+FROM debian:buster
+
+RUN mkdir -p /executable
+WORKDIR /executable
+COPY --from=0 /executable/GRPC_GIT_COMMIT.txt /executable/GRPC_GIT_COMMIT.txt
+COPY --from=0 /executable/bin/worker /executable/bin/worker
+COPY --from=0 /executable/testdata /executable/testdata
```
ybbbby, line 41: Just out of curiosity, what does this testdata do.

ybbbby (approved):

codeblooded (approved):

PR #135, opened 2021-04-25
Delete prototype code.

paulosjca:
Delete prototype code, and apply `go mod tidy`.

Prototype code is not used or supported.


wanlin31:
I don't think I have imported nor observed usage of these code, deleting those should be okay in my opinion, LGTM.

wanlin31 (approved):

codeblooded (approved):

PR #137, opened 2021-04-25
Format markdown files.

paulosjca:


CONTRIBUTING.md
```diff
@@ -36,57 +36,57 @@ from the community if you ask your questions on
...
 
 [golint]: https://github.com/golang/lint
-[Effective Go]: https://golang.org/doc/effective_go.html
-[Code Review Comments]: https://github.com/golang/go/wiki/CodeReviewComments
+[effective go]: https://golang.org/doc/effective_go.html
+[code review comments]: https://github.com/golang/go/wiki/CodeReviewComments
 [go vet]: https://golang.org/cmd/vet/
 [gofmt]: https://blog.golang.org/gofmt
-[Documenting Go Code]: https://blog.golang.org/godoc
+[documenting go code]: https://blog.golang.org/godoc
```
codeblooded, line 92: I've never seen the capitalization of the reference not match its usage in markdown. For example, `[documenting go code]` is the reference defined but it is used as `[Documenting Go Code]`, since it is proper to capitalize articles. It does not appear to be a problem with GitHub Flavored Markdown rendering, but I thought I'd ask if this is desirable? As I said, I have not seen this before.

codeblooded (approved):

wanlin31 (approved):
I don't have much experience regarding the format, take this as a learning opportunity. LGTM

CONTRIBUTING.md
paulosjca: Prettier does that by itself. I think it is desirable, because you may want to capitalize it in one place and not in another when referencing it, but not when defining it. So you might say `The purpose is to replace the [legacy framework]` but in a list you might say `-  [Legacy framework]`, with both referencing the same target defined in the footer.

PR #141, opened 2021-04-30
Remove resource quotas for deployments

codeblooded:
This commit removes the resource quotas for the load test controller and the cleanup agent. The resource quotas were leading to out of memory restarts when batch testing.


wanlin31:
How were you able to find this? I tried to search the logs, did not find any trace, wondering where these information would show up.

paulosjca (approved):

wanlin31 (approved):

ybbbby (approved):

PR #142, opened 2021-04-30
Add regular and prebuilt loadtest example for python_asyncio

ybbbby:
Currently, the regular and prebuilt python_asyncio loadtests would both fail. However, those could be used for bug tracking.

### DIFF
--- a/config/defaults_template.yaml
+++ b/config/defaults_template.yaml
@@ -40,6 +40,10 @@ languages:
     buildImage: l.gcr.io/google/bazel:latest
     runImage: "{{ .ImagePrefix }}python:{{ .Version }}"
 
+  - language: python_asyncio
+    buildImage: l.gcr.io/google/bazel:latest
+    runImage: "{{ .ImagePrefix }}python:{{ .Version }}"
+
   - language: ruby
     buildImage: "{{ .BuildImagePrefix }}ruby:{{ .Version }}"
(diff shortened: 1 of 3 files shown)


codeblooded (approved):

config/defaults_template.yaml
```diff
@@ -40,6 +40,10 @@ languages:
     buildImage: l.gcr.io/google/bazel:latest
     runImage: "{{ .ImagePrefix }}python:{{ .Version }}"
 
+  - language: python_asyncio
+    buildImage: l.gcr.io/google/bazel:latest
+    runImage: "{{ .ImagePrefix }}python:{{ .Version }}"
+
```
wanlin31, line 46: I am not sure if we want to arrange the language alphabetically. :) I think Paulo knows more about this.

wanlin31:
Other than that LGTM.

wanlin31 (approved):

config/defaults_template.yaml
paulosjca: Yes alphabetically please. Thanks.

paulosjca (approved):

config/defaults_template.yaml
ybbbby: I think it is already alphabetical, or did I miss something?

wanlin31: Sorry, I mistakenly think python_Asyncio was in front of python. You are all good.

PR #143, opened 2021-05-01
Add pool label to controller and cleanup agent

codeblooded:
This commit adds an environment variable which sets a pool for the controller and cleanup agent during the deployment.

This is a follow up to #141.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,21 +1,30 @@
 # Version tag for all images
 TEST_INFRA_VERSION ?= "latest"
+
 # Version of the gRPC driver
 DRIVER_VERSION ?= "master"
+
 # Prefix for all images used as clone and ready containers, enabling use with registries
 # other than DockerHub
 INIT_IMAGE_PREFIX ?= ""
+
 # Prefix for all images used as build containers, enabling use with registries
 # other than DockerHub
 BUILD_IMAGE_PREFIX ?= ""
+
 # Prefix for all images used as runtime containers, enabling use with registries
 # other than DockerHub
 IMAGE_PREFIX ?= ""
+
 # Image URL to use all building/pushing image targets
 CONTROLLER_IMG ?= ${IMAGE_PREFIX}controller:${TEST_INFRA_VERSION}
+
 # Image URL to use all building/pushing image targets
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,12 +62,16 @@ deploy: deploy-controller deploy-cleanup-agent
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 deploy-controller: manifests
-	cd config/manager && kustomize edit set image controller=${CONTROLLER_IMG}
+	cd config/manager && \
+		kustomize edit set image controller=${CONTROLLER_IMG} && \
+		kustomize edit set label pool:${SYSTEM_POOL}
```
codeblooded, line 67: @paulosjca Would you rather depend on a specific pool since it is an environment variable in the deployment? Or would you rather just have me set it to go to the nodes with a "default-system-pool" label set to "true"?

paulosjca: The concept of "default pool" is that it is something used by the controller to assign a (client, driver or server) pod to a pool when there is none specified. The actual pod always has a "pool" label set as a node selector.

In the case of the controller itself, there is no entity that would look at the pod configuration and set the  "pool" label, so it has to be set in the deployment step. The code as you have it is correct.

Makefile
```diff
@@ -1,21 +1,30 @@
...
 IMAGE_PREFIX ?= ""
+
 # Image URL to use all building/pushing image targets
 CONTROLLER_IMG ?= ${IMAGE_PREFIX}controller:${TEST_INFRA_VERSION}
+
 # Image URL to use all building/pushing image targets
 CLEAN_IMG ?= ${IMAGE_PREFIX}cleanup:${TEST_INFRA_VERSION}
-#${IMAGE_PREFIX}cleanup_agent:${TEST_INFRA_VERSION}
+
+# Specifies the pool where the controller and cleanup agent must run
```
paulosjca, line 25: Perhaps "Node pool for system components (controller and cleanup agent)", or "Node pool for system components, i.e. controller and cleanup agent"

ybbbby (approved):

paulosjca (approved):

PR #144, opened 2021-05-06
Fix system node selector

codeblooded:
This pull request sets a nodeSelector on the controller and cleanup agent. This selector instructs Kubernetes to place them on nodes with a label of "default-system-pool" set to "true". Unfortunately, kustomize does not allow the nodeSelector field to be edited over the command line. Luckily, this design decouples the name of the actual system pool from the release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,30 +1,21 @@
 # Version tag for all images
 TEST_INFRA_VERSION ?= "latest"
-
 # Version of the gRPC driver
 DRIVER_VERSION ?= "master"
-
 # Prefix for all images used as clone and ready containers, enabling use with registries
 # other than DockerHub
 INIT_IMAGE_PREFIX ?= ""
-
 # Prefix for all images used as build containers, enabling use with registries
 # other than DockerHub
 BUILD_IMAGE_PREFIX ?= ""
-
 # Prefix for all images used as runtime containers, enabling use with registries
 # other than DockerHub
 IMAGE_PREFIX ?= ""
-
 # Image URL to use all building/pushing image targets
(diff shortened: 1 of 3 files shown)


codeblooded:
@paulosjca We cannot set the nodeSelector through kustomize. We'd have to generate a patch file and apply it, which seems like overkill. I went with the other approach, because it was simple and requires no modifications to the deployment process.

Makefile
```diff
@@ -1,30 +1,21 @@
...
 # Image URL to use all building/pushing image targets
 CONTROLLER_IMG ?= ${IMAGE_PREFIX}controller:${TEST_INFRA_VERSION}
-
 # Image URL to use all building/pushing image targets
 CLEAN_IMG ?= ${IMAGE_PREFIX}cleanup:${TEST_INFRA_VERSION}
-
-# Node pool for system components (controller and cleanup agent)
-SYSTEM_POOL ?= "system"
-
+#${IMAGE_PREFIX}cleanup_agent:${TEST_INFRA_VERSION}
```
paulosjca, line 18: Is this comment intended here? Other than that, LGTM.

wanlin31 (approved):
LGTM

wanlin31:
I mean other than Paulo's comment, it looks good to me. :)

ybbbby (approved):

ybbbby:
LGTM, nice investigation!

paulosjca (approved):

Makefile
codeblooded: It was in the previous version on master, so when I reverted... it came back. I can add a commit to remove it

PR #149, opened 2021-05-10
Test runner tool.

paulosjca:
Adds a test runner tool to run multiple sets of tests at a given concurrency level and monitor them to completion.

```
$ ./bin/runner --help
Usage of ./bin/runner:
  -a string
        annotation key to parse for queue assignment (default "pool")
  -c value
        concurrency level, in the form [<queue name>:]<concurrency level>
  -i value
        input files containing load test configurations
  -p duration
        polling interval for load test status (default 20s)
  -retries uint
        Maximum retries in case of communication failure (default 2)
```

For instance, the following command will run tests on two separate queues, as specified by an annotation (`pool: workers-8core` or `pool: workers-32core`), with up to 6 tests from each queue existing at any given time:

```
./bin/runner -i ../grpc/loadtest-8core.yaml -i ../grpc/loadtest-32core.yaml -a pool -c workers-8core:6 -c workers-32core:6
```

The configurations above, containing go tests set up to run on two different testbeds, were generated in a clone of the grpc/grpc repository as follows (with `workers-32core` as an example):

```
tools/run_tests/performance/loadtest_config.py -l go -t ./tools/run_tests/performance/templates/loadtest_template_basic_all_languages.yaml -s client_pool=workers-32core -s server_pool=workers-32core -s big_query_table= -s timeout_seconds=900 --prefix=pcastello -u workers-32core -d --category=scalable -a pool=workers-32core -o ./loadtest-32core.yaml
```
Running tests are polled periodically (with a default interval of 20s) to check for status. (Tests that are initializing are also polled, but half as often.)

The runner currently does not delete tests as they complete. Tests that have a common prefix (which defaults to the user name) can be deleted with `kubectl delete loadtests -l prefix=<prefix>`.

The code in this PR is still missing tests, but these can be added separately.

### DIFF
--- a/tools/runner/client.go
+++ b/tools/runner/client.go
@@ -1,0 +1,70 @@
+package runner
+
+import (
+	"log"
+	"os"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
+	// This side-effect import is required by GKE.
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+
+	grpcv1 "github.com/grpc/test-infra/api/v1"
+	clientset "github.com/grpc/test-infra/clientset"
+)
+
+// NewLoadTestGetter returns a client to interact with LoadTest resources.
+// The client can be used to create, query for status and delete LoadTests.
+func NewLoadTestGetter() clientset.LoadTestGetter {
+	schemebuilder := runtime.NewSchemeBuilder(func(scheme *runtime.Scheme) error {
+		scheme.AddKnownTypes(grpcv1.GroupVersion,
+			&grpcv1.LoadTest{},
+			&grpcv1.LoadTestList{},
+		)
+		metav1.AddToGroupVersion(scheme, grpcv1.GroupVersion)
+		return nil
+	})
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		if err != rest.ErrNotInCluster {
+			log.Fatalf("failed to connect within cluster: %v", err)
+		}
+
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			log.Fatalf("could not find a home directory for user: %v", err)
+		}
+
+		cfgPathBuilder := &strings.Builder{}
+		cfgPathBuilder.WriteString(homeDir)
+		if homeDir[:len(homeDir)-1] != "/" {
+			cfgPathBuilder.WriteString("/")
+		}
+		cfgPathBuilder.WriteString(".kube/config")
+		cfgPath := cfgPathBuilder.String()
+
+		config, err = clientcmd.BuildConfigFromFlags("", cfgPath)
+		if err != nil {
+			log.Fatalf("failed to construct config for path %q: %v", cfgPath, err)
+		}
+	}
+
+	schemebuilder.AddToScheme(clientgoscheme.Scheme)
+	scheme := clientgoscheme.Scheme
+	types := scheme.AllKnownTypes()
+	_ = types
+
+	grpcClientset, err := clientset.NewForConfig(config)
+	if err != nil {
+		log.Fatalf("failed to create a grpc clientset: %v", err)
+	}
+	return grpcClientset.LoadTestV1().LoadTests(corev1.NamespaceDefault)
+}
--- a/tools/runner/runner.go
+++ b/tools/runner/runner.go
@@ -76,40 +76,41 @@ func (r *Runner) Run(qName string, configs []*grpcv1.LoadTest, concurrencyLevel 
 // runTest creates a single LoadTest and monitors it to completion.
 func (r *Runner) runTest(qName string, config *grpcv1.LoadTest, i int, done chan int) {
 	id := fmt.Sprintf("%-14s %3d", qName, i)
+	name := nameString(config)
 	var status string
 	var retries uint
 
 	for {
 		loadTest, err := r.loadTestGetter.Create(config, metav1.CreateOptions{})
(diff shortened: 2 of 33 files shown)


tools/runner/runner.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+// Run runs a set of LoadTests at a given concurrency level.
+func (r *Runner) Run(qName string, configs []*grpcv1.LoadTest, concurrencyLevel int) {
+	done := make(chan int)
+	n := 0
+	for i, config := range configs {
+		if n < concurrencyLevel {
+			go r.runTest(qName, config, i, done)
```
jtattermusch, line 59: qq: will this retry if the API call fails occasionally?

jtattermusch:
nit: I looked at the log and I feel that the log could provide more info for "Running" and "Succeeded" phases.
(e.g. include the name of the loadtest, or some basic details about it's result beyond just "Succeeded". I think that info is going to be useful for debugging. We can of course add this in a followup.
```
2021/05/10 04:57:36 [workers-8core    2] Running
2021/05/10 04:57:36 [workers-32core   2] Running
2021/05/10 04:57:36 [workers-8core    0] Running
2021/05/10 04:57:56 [workers-8core    2] Succeeded
2021/05/10 04:57:56 [workers-32core   2] Succeeded
```

paulosjca:
I have updated the PR to print "reason" and "message" in addition to
"status", and to do so also for tests that are not in running state yet.

That means you will see lines such as this:

2021/05/11 03:30:30 [workers-32core 1] Initializing; PodsMissing; load test
has created 0/3 required pods

Notice that there is no additional info for tests that are "Running" or
"Succeeded".

Run from kokoro;

http://sponge2/8464b5f6-b184-4df3-a69a-30123abde6a1

Thanks,

Paulo

jtattermusch:
>  http://sponge2/8464b5f6-b184-4df3-a69a-30123abde6a1 

Btw, from the log it seems that `[workers-8core    0]` runs for almost 10mins. Initially I didn't understand that at first, but then it I realized it's probably the C++ loadtest and most of the 10mins is actually the build time. Is that correct?
It would be useful if I could see info in the log that would allow me to at least guess that is the case
(e.g. at this point I can't see whether the loadtest is building or actually running) I also have no way of seeing which scenario is the particular load test running (e.g. no way of telling which loadtest is e.g. C++ and what scenario it is running).

paulosjca:
Here is some explanation of message ordering and content:

Log lines are of the form `[<queue name> <test index>] <message>`:

```
2021/05/11 03:29:50 [workers-8core    0] Created test kokoro-test-e943e4f0-f6a3-5df3-b0b5-e281e20ac959
````

The runner enqueues tests for processing in separate queues (determined by an annotation in the test configuration) by creating queue runner goroutines to process the tests assigned each queue. The test index is the index of the test in the list of tests fed to that queue.

Tests from each queue are processed by creating a goroutine to run each particular test. The number of goroutines active at a time is limited to the _concurency level_ for that queue (controlled by the `-c` flag). Every time a test terminates, the queue runner goroutine is notified and another test is allowed to start.
 
The test runner goroutine prints the test name in the "Created" message, but not in subsequent messages, in order to keep the log more readable. You can find the name of the test by finding the creation message for that queue and index.

After the test is created, it is polled periodically and the status is logged. Once the test terminates, that is also logged one last time before the test runner goroutine returns.

This may help to explain why messages for different `[<queue name> <index>]` combinations do not appear in a set order.

The scenario executed by each test is not printed by the runner, but is present as an annotation in the test configuration. The index is determined by the position of the configuration in the input file, so this will print all scenario names in order (replace `loadtest-8core.yaml` with any configuration file containing tests for a single queue):

```
grep '^    scenario:' loadtest-8core.yaml
```

Configuration generation is deterministic, so you can reproduce the same files used continuous run by running the same command line. in this particular case:

```
tools/run_tests/performance/loadtest_config.py -l c++ -l go \
    -t ./tools/run_tests/performance/templates/loadtest_template_basic_all_languages.yaml \
    -s client_pool=workers-8core -s server_pool=workers-8core \
    -s big_query_table= -a pool=workers-8core -s timeout_seconds=900 \
    --prefix="kokoro-test" -u "$(date +%Y%m%d%H%M%S)" -u workers-8core \
    -r '(go_generic_sync_streaming_ping_pong_secure|go_protobuf_sync_unary_ping_pong_secure|cpp_protobuf_async_streaming_qps_unconstrained_secure)$' \
    -o ./loadtest-8core.yaml
```

**Comment by paulosjca** [comment]
I have updated the PR to show a  "readable" name in addition to the name with a UUID:

```
2021/05/11 11:42:31 [workers-32core   1] Created test kokoro-test-go-generic-sync-streaming-ping-pong-secure-20210511114229-workers-32core [kokoro-test-70e2a110-48fd-5b6b-92a6-fa4b6ed667b2]
```

We still need the name with a UUID to locate the test on the cluster.

The reason why we use a name with a UUID is because the controller implementation adds the LoadTest **name** as a label to the pods, so it can refer back to the LoadTest from the pods. This introduces a limitation in the names, since labels are limited to 63 characters (while names themselves can go up to 253 characters). LoadTests also have a **UID** that could be used instead of the name . Once we make that change, we can use the "readable" name as the LoadTest name in the resource metadata.

File: tools/runner/client.go [hunk 2]
```diff
@@ -0,0 +1,70 @@
...
+			&grpcv1.LoadTestList{},
+		)
+		metav1.AddToGroupVersion(scheme, grpcv1.GroupVersion)
+		return nil
+	})
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		if err != rest.ErrNotInCluster {
+			log.Fatalf("failed to connect within cluster: %v", err)
```
wanlin31, line 38: I am a bit confused, in this case do we regard the message in log.Fatalf as error message or information message?

paulosjca: This code came from the example under cmd/clientset. We should improve it to not use log.Fatalf (it should probably return an error instead of exiting), but in this case it is only caused once from main(), so the effect is the same. I would like to defer looking at this until after we merge this PR.

wanlin31 (approved):
LGTM!

PR #150, opened 2021-05-11
Feature/parallel building

wanlin31:
This commits replaced the sequencial image building to parallel build, this update reduced the total image building time to 19 mins from Kokoro VM. 

This commits has also add unique identifier for each image build to prevent cache of old repository. Also add build-only option.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/go-logr/logr v0.1.0
 	github.com/golang/protobuf v1.4.2
 	github.com/google/uuid v1.1.1
+	github.com/nu7hatch/gouuid v0.0.0-20131221200532-179d4d0c4d8d // indirect
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
 	github.com/pkg/errors v0.8.1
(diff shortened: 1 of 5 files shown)


wanlin31:
Just realized that we have several languages building from same repo: grpc/grpc, so to break the cache we need to give a unique identified each time. Fixing it now.

ybbbby:
Since most of time we would build all those languages together. Storing grpc/grpc in cache sounds like a good thing? > Just realized that we have several languages building from same repo: grpc/grpc, so to break the cache we need to give a unique identified each time. Fixing it now.

wanlin31:
> Since most of the time we would build all those languages together, storing grpc/grpc in cache sounds like a good thing? > Just realized that we have several languages building from same repo: grpc/grpc, so to break the cache we need to give a unique identified each time. Fixing it now.

I think your suggestion makes scene, thanks for reminding me! 

For continuous build, we are checking out specific commit sha, using cashed core repo should not hurt.

For the manual users, each round of building would be different, we would only care about the cache within each round of building. I think we actually should freeze to a certain version of the repo as soon as we start building, since our building time is 19 mins or so, it is possible that someone makes a change and pushed to the repository, and if the user only give something like cxx:master, the source code would be changed, the workers would not be building from the same snapshot without knowing it. 

I would use a time when we start to build the image as cache breaker so that each round the repo would be fresh, but within each building round, we could use the cached repo.

 I might have over complicated the situation, would love hear thoughts from @codeblooded and @paulosjca

paulosjca (approved):

PR #159, opened 2021-05-15
Move log messages and reporting out of the test runner.

paulosjca:
This change moves message logging out of the test runner, and makes several interface changes to clean up the flow of the runner and prepare for report generation.


tools/runner/reporter.go
```diff
@@ -0,0 +1,102 @@
...
+
+// StartTest records the start of the test and its start time.
+func (r *TestCaseReporter) StartTest(startTime time.Time) {
+	r.StartTime = startTime
+}
+
+// EndTest records the end of the test and its end time.
+func (r *TestCaseReporter) EndTest(endTime time.Time) {
+	r.Duration = endTime.Sub(r.StartTime)
+}
```
codeblooded, line 97: nit: Perhaps, do we need to inject the times? These functions feel like setters for attributes which are already exposed on the struct.

From my perspective, several assumptions are made:

- When the name is StartTest, an assumption is that the test starts with the call. For this reason, I think we can use `r.StartTime = time.Now()`.

- When the name is EndTest, an assumption is likewise made that the test ends with the call. To simplify the logic here, we can replace `r.Duration = time.Since(r.StartTime)`.

If the goal is to mock time for testing, there are some other ways we can achieve that. We can define a helper and override it to set the time to the proper point. That said, I think the majority of tests for the reporter would be validating the XML fields, not strings.

tools/runner/reporter.go
```diff
@@ -0,0 +1,102 @@
...
+
+// TestDuration returns the duration of the test.
+func (r *TestCaseReporter) TestDuration() time.Duration {
+	return r.Duration
+}
```
codeblooded, line 102: If we plan to have getters, I think `Duration` and `StartTime` should not be exported.

paulosjca: Done. These are placeholders in any case. Added TODOs to reflect that.

codeblooded (approved):

PR #160, opened 2021-05-19
Log details of Kubernetes API errors in the runner

codeblooded:
This commit adds Kubernetes API error details to the log output of the runner, allowing gRPC developers to see what went wrong. This can aid in debugging.


paulosjca (approved):
Thanks. We typically run on go 1.12 on kokoro, so we can't do much better than adding the error message here.

wanlin31 (approved):

ybbbby (approved):

PR #161, opened 2021-05-19
Update README.md

wanlin31:
Some cleanup of README for image builder and image deletion.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -42,7 +42,7 @@ May be repeated.
 The script [delete_prebuilt_workers.go](prebuilt_workers/delete_prebuilt_workers.go) 
 deletes images within user specified registry. The script lists all images
 within the specified registry, then check if the image has the user specified 
-tag. The script only support Google cloud registry so far.
+tag. This script only supports Google Container Registry, because it relies on the google-cloud-sdk.
 
 The following example delete all images within 
 `gcr.io/grpc-testing/project-name/pre_built_workers` that have 
--- a/tools/README.md
+++ b/tools/README.md
@@ -23,7 +23,7 @@ to test.
 
 The script `prepare_for_prebuilt_workers.go` takes the following options:
 * `-l `<br> Language and GITREF to benchmark. The language and its specific 
-GITREF wish to build workers from can be specified as `language:COMMIT_SHA`.
+GITREF to build the workers from can be specified as `language:COMMIT_SHA`.
 May be repeated.
 * `-t` <br> Tag for prebuilt images. Tag is a required fields. Tag complies with 
 [docker tag's restrictions](https://docs.docker.com/engine/reference/commandline/tag/#extended-description). 
--- a/tools/README.md
+++ b/tools/README.md
@@ -5,14 +5,10 @@ The scripts in this folder:
 * Push the images to specified registry.
 * Delete the images from specified registry.
 
-The helper scripts here should be specifically used with GCR. The predominate drive of developing these
-scripts are to assist continuous build. The scripts could also help with manual usage
-of prebuilt images if working with GCR. If another image registry is chosen, the prebuilt workers Dockerfiles are available. The root directory of the Dockerfiles could be find [test-infra/containers/pre_built_images](test-infra/containers/pre_built_images).
-
 ## Build and push images
 
 The script [prepare_prebuilt_workers.go](pre_built_workers/prepare_prebuilt_workers.go) 
-builds images and pushes them to a user specified registry. For example, the following shows this process (building and pushing prebuilt images) for cxx and go workers:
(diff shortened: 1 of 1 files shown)


wanlin31:
I opened this PR to follow up on cleanup items we talked about before merging PR113. (image building tool).

ybbbby (approved):

tools/README.md
```diff
@@ -27,16 +23,15 @@ to test.
 
 The script `prepare_for_prebuilt_workers.go` takes the following options:
 * `-l `<br> Language and GITREF to benchmark. The language and its specific 
-GITREF wish to build workers from can be specified as `language:gitref`.
+GITREF wish to build workers from can be specified as `language:COMMIT_SHA`.
```
codeblooded, line 26: Nit: I think the word "wish" makes it sound like we're introducing a new clause. We can just omit it. "from" gets the point across. 

```suggestion
GITREF to build the workers from can be specified as `language:COMMIT_SHA`.
```

tools/README.md
```diff
@@ -27,16 +23,15 @@ to test.
...
-  `-build-only=true` images will only be built locally without being pushed to an 
-  image registry.
-* `-p` <br> Image registry to store images. In the example, the location of the
-images are:
+  `-build-only=true`, images will only be built locally.
+* `-p` <br> Image registry to store images. The helper script only support pushing images to GCR. If another image registry is chosen, the images could still be built through script with `-build-only=true`, the user could then push the images manually. The Dockerfiles that the script building from is avalible in [test-infra/containers/pre_built_images](test-infra/containers/pre_built_images). If using Google cloud registry, the image prefix should be in form of `gcr.io/project-name/your-directory-name`. In the example, the built images would be stored as: 
+ :
   ```
   gcr.io/grpc-testing/project-name/pre_built_workers/cxx:user-specified-tag
   gcr.io/grpc-testing/project-name/pre_built_workers/go:user-specified-tag
```
codeblooded, line 42: This flow is a bit confusing. Let's try to break it up a bit. 

```suggestion
* `-p` <br> Image registry to store images. The image prefix should be in form of `gcr.io/project-name/your-directory-name` for the Google Container Registry. For example, the built images would be stored as `gcr.io/grpc-testing/project-name/pre_built_workers/cxx:user-specified-tag` and `gcr.io/grpc-testing/project-name/pre_built_workers/go:user-specified-tag`. 
   
   If using a registry other than GCR, the images should be built through the script with the flag `-build-only=true`. The user could then push the images manually.
   
   The Dockerfiles that the script uses to build are available in [test-infra/containers/pre_built_images](test-infra/containers/pre_built_images).
```

As for paragraph 2, "user could the push the images manually", this paragraph may be omitted. If the command is `docker push` and they set the image prefix to the appropriate path, it will work for any registry. For example,

- Dockerhub: set prefix to "username/" then `docker push username/image` will work.

- AWS registries: set prefix to their structure `aws-account-id.dkr.ecr.region.amazonaws.com/image` then `docker push aws-account-id.dkr.ecr.region.amazonaws.com/image` works the same way. So does Microsoft Azure registries, I believe.

Unless we revert to gcloud commands, there shouldn't be a need to make them perform extra steps.

codeblooded: If we need multiple paragraphs, I'd remove the `<br>` statements. We can hold the flags for emphasis. HTML tags in markdown aren't the cleanest. It's generally best to avoid them whenever possible.

tools/README.md
```diff
@@ -47,7 +42,7 @@ images are:
 The script [delete_prebuilt_workers.go](prebuilt_workers/delete_prebuilt_workers.go) 
 deletes images within user specified registry. The script lists all images
 within the specified registry, then check if the image has the user specified 
-tag.
+tag. The script only support Google cloud registry so far.
```
codeblooded, line 45: A couple comments:

- We don't have any intention to support other registries with this script right now, and we're not really looking for PRs with them. Therefore, I'd just leave off the "so far" phrase. 

- The official name of the product is "Google Container Registry," so let's just use that. You can add that the limitation is due to depending on the google-cloud-sdk. 
 
```suggestion
tag. This script only supports Google Container Registry, because it relies on the google-cloud-sdk.
```

codeblooded (changes requested):
Looking very good Wanlin. I've included a few points and suggestions.

tools/README.md
paulosjca: The <br> is very useful for legibility. Please keep it.

paulosjca (approved):
We can fix it again later as needed.

tools/README.md
wanlin31: I think it is because I update README from UI, it doesn't allow me to modify it from remote client. I am closing this PR and open a new one.

PR #163, opened 2021-05-20
Add LoadTest state information to kubectl get

codeblooded:
This commit uses fields on the Custom Resource Definition to customize the behavior of the `kubectl get loadtest[s]` command. It now shows the state of the test if it has started or completed. This allows a developer to see the test's progress without having to `kubectl describe` it.

### DIFF
--- a/api/v1/loadtest_types.go
+++ b/api/v1/loadtest_types.go
@@ -320,7 +320,7 @@ type LoadTestSpec struct {
 // components. If any one component has errored, the load test will be marked in
 // an Errored state, too. This will occur even if the other components are
 // running or succeeded.
-// +kubebuilder:default=Unrecognized
+// +kubebuilder:default=Unknown
 type LoadTestState string
 
 const (
@@ -416,6 +416,8 @@ type LoadTestStatus struct {
 // +kubebuilder:subresource:status
 
 // LoadTest is the Schema for the loadtests API
+// +kubebuilder:printcolumn:name="State",type=string,JSONPath=`.status.state`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 type LoadTest struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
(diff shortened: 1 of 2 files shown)


codeblooded:
A small change that we can extend to provide more information to developers. I think it works towards fixing something easy in @drfloob's friction log.

config/crd/bases/e2etest.grpc.io_loadtests.yaml
```diff
@@ -8,6 +8,13 @@ metadata:
   creationTimestamp: null
   name: loadtests.e2etest.grpc.io
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .metadata.creationTimestamp
```
paulosjca, line 12: Pods display "Age" last, so perhaps we should have State before Age for the load tests as well.

Generally speaking, I am more interested in whether a test has an error than whether it is 9h or 10h old.

api/v1/loadtest_types.go
```diff
@@ -320,7 +320,7 @@ type LoadTestSpec struct {
 // components. If any one component has errored, the load test will be marked in
 // an Errored state, too. This will occur even if the other components are
 // running or succeeded.
-// +kubebuilder:default=Unrecognized
+// +kubebuilder:default=Unknown
```
wanlin31, line 323: Could you provide reference for usage of `+kubebuilder:default=`, I want to learn more about markers however could not find example of using `+kubebuilder:default=`. Thanks.

codeblooded: The `+kubebuilder:default=` marker has no examples. The documentation is virtually nothing more than its signature: https://kubebuilder.io/reference/markers/crd-validation.html#:~:text=default. Defaults will not work unless we upgrade to Kubebuilder v3. We need a bug to do that anyway.

wanlin31: Thank you for pointing the reference, wondering what is the difference for Unrecognized and unknow?

wanlin31 (approved):

ybbbby (approved):

config/crd/bases/e2etest.grpc.io_loadtests.yaml
paulosjca: I am curious because the additionalPrinterColumns still show Age first. Is it going to display Age last, with the PR as it is?

paulosjca (changes requested):

paulosjca (approved):

PR #164, opened 2021-05-24
Use "php7" language name in defaults and examples.

paulosjca:
This change renames the language in the defaults and examples from "php" to "php7", to match the language name in test scenarios. This change requires a new controller release before default images can be used. Tests with prebuilt images are not affected.

### DIFF
--- a/tools/prepare_prebuilt_workers/prepare_prebuilt_workers.go
+++ b/tools/prepare_prebuilt_workers/prepare_prebuilt_workers.go
@@ -107,7 +107,7 @@ func main() {
 		split := strings.Split(pair, ":")
 
 		if len(split) != 2 || split[len(split)-1] == "" {
-			log.Fatalf("Input error in language and gitref selection, please follow the format as language:gitref, for example: cxx:<commit-sha>")
+			log.Fatalf("Input error in language and gitref selection, please follow the format as language:gitref, for example: c++:<commit-sha>")
 		}
 
 		lang := split[0]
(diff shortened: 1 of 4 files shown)


wanlin31 (approved):
I think there are other places we have php instead of php7, I will create anther PR to update names from php-php7

tools/prepare_prebuilt_workers/prepare_prebuilt_workers.go
```diff
@@ -107,7 +107,7 @@ func main() {
 		split := strings.Split(pair, ":")
 
 		if len(split) != 2 || split[len(split)-1] == "" {
-			log.Fatalf("Input error in language and gitref selection, please follow the format as language:gitref, for example: cxx:<commit-sha>")
+			log.Fatalf("Input error in language and gitref selection, please follow the format as language:gitref, for example: c++:<commit-sha>")
```
wanlin31, line 110: Are we trying update cxx to c++? If so I will create another PR to update everything from cxx to c++. I think if we only update it here, it might cause some script to not work.

paulosjca: We cannot update cxx to c++ because the character "+" cannot be used in labels. This is the only language we will always need to translate. But we accept c++ in the value of flags, and we should, because that is the language name used in scenario generation.

wanlin31: Right! I forgot you test cxx before, so in this case everything in test-infra stays as `cxx`.

PR #165, opened 2021-05-24
Update php to php7 in test-infra for remaining places

wanlin31:
This commit follows PR# 164 to update php to php7 at remaining
places.


paulosjca (approved):
Thanks, I must have missed that in the PR, although I thought I had included it.

PR #166, opened 2021-05-26
DO NOT MERGE: Separates junit reports from reporter.

paulosjca:


codeblooded:
Closing since #157 is merged with a similar approach.

PR #169, opened 2021-05-26
Delete references to label "loadtest" from controller.

ybbbby:
Also, added ownerReferences to pod in several scenarios in load_controller_test.

### DIFF
--- a/status/relationships.go
+++ b/status/relationships.go
@@ -20,7 +20,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	grpcv1 "github.com/grpc/test-infra/api/v1"
-	"github.com/grpc/test-infra/config"
 )
 
 // PodsForLoadTest returns a slice of pointers to pods which belong to a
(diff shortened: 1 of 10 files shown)


paulosjca (approved):

status/relationships.go
```diff
@@ -30,15 +29,16 @@ func PodsForLoadTest(loadtest *grpcv1.LoadTest, allPods []corev1.Pod) []*corev1.
...
 	for i := range allPods {
 		pod := &allPods[i]
 
-		parent, ok := pod.Labels[config.LoadTestLabel]
-		if ok && parent == loadtest.Name {
-			pods = append(pods, pod)
+		for _, owner := range pod.GetOwnerReferences() {
+			if owner.UID == loadtest.GetUID() {
+				pods = append(pods, pod)
+				break
```
wanlin31, line 40: Why do we break here? I thought we need to get a list of the pods related to the owner.

wanlin31 (commented):
Looks good to me, just have a minor question. :)

status/relationships.go
paulosjca: The logic here is the opposite: it is iterating on pods to find which pods belong to a given load test. Once it finds that the load test is in the pod's list of owners, it is done with the pod.

ybbbby: Yeah, I agree this break is kind redundant here. For now, each pod would only get one owner which is the loadtest. So in theory this for loop would only execute once anyway. However, if for some reason, one pod could have multiple owners, once we found this pod belongs to this loadtest, we could jump out of inner loop before finishing iterating all owners to save time. Does it make sense?

paulosjca: I am pessimistic, so I am thinking of the possibility that there might be repeated UIDs in the owners list :) I think it is more elegant to break once the pod is appended, because this is an action that should only happen once, and we don't have to depend for that on the exact behavior of GetOwnerReferences().

PR #172, opened 2021-06-02
Feature/add driver timeout

wanlin31:
This commit adds timeout for driver pods. The environment variables: POD_TIMEOUT and KILL_AFTER are added at the time when the pods are created, this step is similar to all pods (driver, client and servers). POD_TIMEOUT takes the value of TimeoutSecond, which is accessible through individual LoadTest yaml configuration. KILL_AFTER could be set through defaults.yaml at the time of deployment. For driver the timeout is directly assigned to running the driver.


config/cmd/configure.go
```diff
@@ -41,6 +41,8 @@ type DefaultsData struct {
 	InitImagePrefix  string
 	ImagePrefix      string
 	BuildImagePrefix string
+	PodTimeout       string
```
paulosjca, line 44: The timeout should not be specified in defaults, but should be derived from the test timeout. I think it does make sense to specify "KillAfter" as some sort of global default.

paulosjca: I think it might be better to make PodTimeout a structure in the LoadTest configuration, with fields like GracePeriodSeconds and KillAfterSeconds. The first one would be added to the test timeout to create the pod timeout. In any case, we should consider the interaction between the test timeout and the pod timeout – do we want a race between the pod and the test timing out? would that affect the end state? We should give it some thought.

wanlin31: Right, I create this PR to collect thoughts on how much flexibility we want on Podtimeout and Killafter.  I agree that Killafter is somewhat default that most user doesn't have to touch, it is  just to make sure that the timeout is executed. I can see the  Podtimeout time to be set through LoadTest configuration files, since it has the similar effect as TimeoutSeconds.

The Podtimeout count start the moment when the run container starts, it protects the case that the pod is successfully get to run container but the process within run container went wrong. 

Timeoutseconds tries to guard the total amount of time starting from the LoadTest being scheduled, but the only thing it does is to mark the LoadTest as TimeoutErrored and send asynchronous RPC to workers, there is no guarantee the workers will be stopped.  (1) With scenarios like the building process went wrong (not using prebuild image), if initial containers take longer than Timeoutseconds, the LoadTest will be marked as TimeoutErrored, but the workers will ignore the RPC. The behavior of Timeoutseconds is similar to not any control. (2) If the initial process went through okay, the process in run containers goes wrong, TimououtSeconds may be able stop the workers, which is also similar to what Podtimeout could achieve for sure.

I am thinking of using Podtimeout to replace Timeoutseconds, and having driver and workers have same PodtimeoutSeconds, as Ready container timeout was set as Timeoutseconds, we may want to set it with PodtimeoutSeconds to give enough time for workers to be up and running. 

It should work like this:
With prebuilt image 
(1) Workers up and running okay --> Driver is up --> Driver Timeout --> LoadTest Marked as Timeout Error --> Workers timeout on their own
(2) Workers up and running okay --> Driver is up --> Worker Timeout --> LoadTest Marked as Timeout Error --> Driver timeout 
(3) Workers could not run (stuck before get to run container)--> Driver Ready Container Timeout -->  LoadTest Marked as Timeout Error, in this case the worker might stuck waiting for TTL seconds to delete the pods all together, which is what we have with  Timeoutseconds.

Just my thoughts, we could discuss further in the meeting. :)

config/defaults.go
```diff
@@ -80,6 +86,11 @@ func (d *Defaults) Validate() error {
 		}
 	}
 
+	if d.KillAfter == "" {
+		fmt.Println("KillAfter time is not set-up default to 15s")
+		d.KillAfter = "15s"
+	}
+
```
wanlin31, line 93: I think we should make KillAfter a required field instead of giving it a default time. Leave it here just want to get more opinion.

podbuilder/podbuilder.go
```diff
@@ -310,6 +311,14 @@ func (pb *PodBuilder) newPod() *corev1.Pod {
...
 	}
 
+	pb.run.Env = append(pb.run.Env, corev1.EnvVar{
+		Name:  "POD_TIMEOUT",
+		Value: strconv.FormatInt(int64(pb.test.Spec.TimeoutSeconds), 10)})
+
+	pb.run.Env = append(pb.run.Env, corev1.EnvVar{
+		Name:  config.KillAfter,
+		Value: pb.defaults.KillAfter})
+
```
wanlin31, line 321: I add the environment variables here since we would also want to have similar features for workers.

config/cmd/configure.go
```diff
@@ -89,6 +90,8 @@ container images that are not used as init containers.`)
 
 	flag.BoolVar(&validate, "validate", true, "validate the output configuration for correctness")
 
+	flag.StringVar(&data.KillAfter, "kill-after", "", "time allowed for pod to response")
```
paulosjca, line 93: Perhaps "time allowed for pod to respond after timeout"

config/constants.go
```diff
@@ -110,4 +110,7 @@ const (
 	// WorkspaceVolumeName contains the name of the volume that is shared between
 	// the init containers and containers for a driver or worker pod.
 	WorkspaceVolumeName = "workspace"
+
+	// KillAfter sets the allowed response time for a pod
```
paulosjca, line 114: Here also "sets the allowed time for a pod to respond after timeout." This one should have a period at the end.

config/defaults_test.go
```diff
@@ -54,6 +54,7 @@ var _ = Describe("Defaults", func() {
 					RunImage:   "gcr.io/grpc-fake-project/test-infra/java",
 				},
 			},
+			KillAfter: "20",
```
paulosjca, line 57: Perhaps we want to validate the format of KillAfter. Perhaps it should be of type time.Duration, and converted to seconds for the environment variables. Or perhaps it should always be a number representing the duration in seconds, to keep it simple.

If we choose to use the "number of seconds", perhaps we should call it KillAfterSeconds.

podbuilder/podbuilder.go
```diff
@@ -310,6 +311,14 @@ func (pb *PodBuilder) newPod() *corev1.Pod {
 		})
 	}
 
+	pb.run.Env = append(pb.run.Env, corev1.EnvVar{
+		Name:  "POD_TIMEOUT",
+		Value: strconv.FormatInt(int64(pb.test.Spec.TimeoutSeconds), 10)})
```
ybbbby, line 316: Does it mean the pod would timeout after 10s?

paulosjca: > Does it mean the pod would timeout after 10s?

Good point. strconv.Itoa() might be less confusing here.

paulosjca: Or perhaps just fmt.Sprint(), which would bypass the conversion to int64.

config/defaults.go
```diff
@@ -48,6 +49,9 @@ type Defaults struct {
 	// Languages specifies the default build and run container images
 	// for each known language.
 	Languages []LanguageDefault `json:"languages,omitempty"`
+
+	// KillAfter specifies the allowed time for a time to respond to a singnal
```
codeblooded, line 53: typo:

```suggestion
	// KillAfter specifies the allowed time for a time to respond to a signal
```

config/cmd/configure.go
codeblooded, line 93: `KillAfter` is a duration in the format described by the time package. Maybe we should parse it as a time.Duration with `flag.DurationVar(&data.KillAfter, ...)`. I understand that it must be passed as a string to the environment variable, but there is no validation. You can use a time.Duration and convert back to a string with https://pkg.go.dev/time#Duration.String.

Consider this scenario: the user passes a bad duration string. For example, "400a" instead of "400s". If we just use a string, we won't know this is problematic with the deployment configuration itself. We will only find out after we have deployed to the cluster and start running tests.

I agree `string -> time.Duration -> string` appears redundant, but the first conversion occurs under-the-hood. And, if it provides validation... I think that is an improvement. The other choice would be to manually invoke time.ParseDuration and error check. But, using the duration flag as intended feels less hacky to me.

podbuilder/podbuilder.go
codeblooded: +1 to @paulosjca feedback. I think `fmt.Sprintf("%d", pb.test.Spec.TimeoutSeconds)` is cleanest.

config/cmd/configure.go
codeblooded: I see @paulosjca covered this in https://github.com/grpc/test-infra/pull/172/files#r645650206 as well. I am in agreement. Since Go allows Durations as a native type, I think we should utilize them. Interestingly, the `time.ParseDuration` function grammar is very similar to the `timeout --killafter` grammar if not identical.

codeblooded (changes requested):

config/defaults_test.go
codeblooded: If it is # of seconds, let's make it an integer and not a string. I'd prefer time.Duration... but I'm okay with integer seconds. I think string seconds is a tad confusing. The timeout command's documentation says it supports the same suffixes as Go, so (time.Duration).String should give the right format if supplied seconds, minutes, hours or days. https://man7.org/linux/man-pages/man1/timeout.1.html

podbuilder/podbuilder.go
wanlin31: Thank you! I was not sure which to use, thank you for helping here. :)

config/defaults.go
wanlin31: I rewrite this sentence, thank you for pointing this out!

config/cmd/configure.go
wanlin31: You are making a great point! I have learned that we should always make some effort if possible to validate input information as early as possible, I should not trust the input too much. I realized the validation is really important!

config/defaults_test.go
wanlin31: I named it as KillAfterSeconds, so that we can use integer second. I decided to use int64 so that we would need to introduce new package. Is there any difference in terms of using int64 or duration in the respect of validation?

paulosjca: I think that, since it is a duration, it is less confusing to actually make it of type Duration. Then we can also just call it "KillAfter" etc, without the seconds. When we create the KILL_AFTER environment variable, at this point I would convert to .Seconds() to avoid issues (a duration of "30m20s" is valid in Go but not as an argument to "timeout").

config/constants.go
```diff
@@ -110,4 +110,7 @@ const (
 	// WorkspaceVolumeName contains the name of the volume that is shared between
 	// the init containers and containers for a driver or worker pod.
 	WorkspaceVolumeName = "workspace"
+
+	// KillAfter sets the allowed response time for a pod after timeout.
+	KillAfterSeconds = "KILL_AFTER_SECONDS"
```
paulosjca, line 115: I think we want "KILL_AFTER" here.

wanlin31: I forgot the modify KILL_AFTER to KILL_AFTER_SECONDS earlier in the run.sh file, sorry for the confusion. I think it makes scene to me o be consistent.

wanlin31:
It was a branch issue that causing me not seeing the typos from `make test`, now after I fixed the branch issue, I saw some errors come out, fixing it, sorry for the inconvenience.

config/defaults_test.go
wanlin31: Make scene, I will make it less confusing by using duration instead of integer.

config/defaults.go
```diff
@@ -80,6 +87,10 @@ func (d *Defaults) Validate() error {
 		}
 	}
 
+	if d.KillAfter == -1 {
+		return errors.Errorf("killAfter missing")
+	}
+
```
wanlin31, line 93: For flag passing a duration, I don't have the chance to leave it empty, I have to give it some default value, I set up the default value to be negative to make sure we catch the error. Is this a too hacky?

paulosjca: Can't it be zero, or something like that? The question would be if we want to always require a 'killafter', or not. The behavior when killafter of 0 is passed to timeout is to not implement killafter. So I think 0 would have the correct behavior.

paulosjca: But I see your point, it should be set to something and we probably don't want to let the user avoid setting it. So setting it to -1 sounds ok to me.

paulosjca (approved):

wanlin31:
I will do a final round of testing, and merge this PR to fulfill the CI KR.

wanlin31:
During testing, I realized that if parse killAfter as duration, the default unit is ns and only taking numbers, for example : killAfter: 20 means 20ns, if we want it to be 20s, the value will be 20000000000, 20s would not be accepted.

I am not aware any other way to make it nicer when we parse it directly to duration, is it okay to have it set up this way? 

Or I can change it to a string and use time.ParseDuration(string) to convert it to duration immediate after parsing, we can have  validation when we parse the string to duration.

paulosjca:
That is a good reason to make KillAfter be of type float64 (since in the end we are going to use it as a float), and just put in the help string that it is "in seconds".

config/defaults.go
paulosjca: > That is a good reason to make KillAfter be of type float64 (since in the end we are going to use it as a float), and just put in the help string that it is "in seconds".

For the default value, you could use `math.NaN()`, and then check if it was set with `math.IsNaN(d.KillAfter)`.

paulosjca: Actually we should check the flag itself, data.KillAfter, with math.IsNan(), and perhaps also that it is not negative.  For d.KillAfter, we should just check that it is not negative perhaps.

wanlin31: I think at the time we pass a values through flag, in configure.go file, missing input was checked with `flag.NArg() != 2`, it only allows 2 flag with no argument input, might be sufficient for missing input check? I added the check for d.KillAfter to be non-negative

paulosjca: I tested the duration and it turns out "10" gives a parsing error, so I think we are safe in using Duration for the flag. The default value should probably be something like -time.Second (or -1 * time.Second if we want to be explicit).

paulosjca: Let's go with what is there now, because it works and the type of the flag in configure.go is a minor implementation detail, not an interface.

paulosjca (approved):

config/defaults.go
```diff
@@ -80,6 +86,10 @@ func (d *Defaults) Validate() error {
 		}
 	}
 
+	if d.KillAfter < 0 {
```
paulosjca, line 89: Should update the error message to be more explicit "killAfter must not be negative".

config/cmd/configure.go
```diff
@@ -89,6 +91,8 @@ container images that are not used as init containers.`)
 
 	flag.BoolVar(&validate, "validate", true, "validate the output configuration for correctness")
 
+	flag.Float64Var(&data.KillAfter, "kill-after", math.NaN(), "time allowed for pod to response after timeout, the value should be in seconds")
+
 	flag.Parse()
 
 	if flag.NArg() != 2 {
```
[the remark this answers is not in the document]
paulosjca: Here we need to check the value with math.IsNan(). The flag.NArg() != 2 is to check if there are two arguments left (input and output), but that will happen whether you set the kill-after flag or not.

paulosjca (changes requested):

config/defaults.go
wanlin31: I tried the duration, "10s" gives a parsing error too, it is actually more convenient to use float64 than using Duration. I will just kept it with float64. Thank you for testing this with me.

config/cmd/configure.go
```diff
@@ -41,6 +42,7 @@ type DefaultsData struct {
 	InitImagePrefix  string
 	ImagePrefix      string
 	BuildImagePrefix string
+	KillAfter        float64
```
jtattermusch, line 45: nit: it seems like killafter should be an integer?

config/cmd/configure.go
```diff
@@ -89,12 +91,18 @@ container images that are not used as init containers.`)
 
 	flag.BoolVar(&validate, "validate", true, "validate the output configuration for correctness")
 
+	flag.Float64Var(&data.KillAfter, "kill-after", math.NaN(), "time allowed for pod to response after timeout, the value should be in seconds")
```
jtattermusch, line 94: nit: s/to response/to respond/

config/constants.go
```diff
@@ -110,4 +110,7 @@ const (
 	// WorkspaceVolumeName contains the name of the volume that is shared between
 	// the init containers and containers for a driver or worker pod.
 	WorkspaceVolumeName = "workspace"
+
+	// KillAfter sets the allowed response time for a pod after timeout.
+	KillAfter = "KILL_AFTER"
```
jtattermusch, line 115: nit: name this `KillAfterEnv` to match the style of already-existing `ScenariosFileEnv`?

jtattermusch (commented):
I skimmed through the changes and overall it looks good.
Left a few comments. I'll let someone else approve after reviewing in detail.

config/cmd/configure.go
```diff
@@ -89,12 +91,18 @@ container images that are not used as init containers.`)
...
+
 	flag.Parse()
 
 	if flag.NArg() != 2 {
 		exitWithErrorf(1, true, "missing required arguments")
 	}
 
+	if math.IsNaN(data.KillAfter) {
+		exitWithErrorf(1, true, "missing kill-after argument")
```
paulosjca, line 103: Nit: Perhaps "missing required flag: kill-after".

config/defaults.go
```diff
@@ -80,6 +86,10 @@ func (d *Defaults) Validate() error {
...
+		return errors.Errorf(fmt.Sprintf("killAfter must not be negative, input value is: %f", d.KillAfter))
```
paulosjca, line 90: Nit: perhaps remove ", input value is".

paulosjca (approved):

config/cmd/configure.go
paulosjca: I think it is a minor issue whether it is one or the other, but the kill-after flag in timeout is a float, so there is no harm in allowing it to be specified as a float here.

paulosjca:
It looks like we need approval from @codeblooded on the requested changes.

config/defaults.go
```diff
@@ -80,6 +86,10 @@ func (d *Defaults) Validate() error {
 		}
 	}
 
+	if d.KillAfter < 0 {
+		return errors.Errorf("killAfter must not be negative")
```
codeblooded, line 90: This provides validation at runtime that the KillAfter field is not negative. Should there also be validation before deploying and starting the controller (at configure time)?

config/defaults.go
```diff
@@ -48,6 +51,9 @@ type Defaults struct {
 	// Languages specifies the default build and run container images
 	// for each known language.
 	Languages []LanguageDefault `json:"languages,omitempty"`
+
+	// KillAfter is the duration allowed for pods to respond after timeout.
+	KillAfter float64 `json:"killAfter"`
```
codeblooded, line 56: We can leave it as "KillAfter" in the code, but perhaps we should consider making it "killAfterSeconds" in the JSON with `json:"killAfterSeconds"`. Without the unit, I don't know what `KillAfter: 20` means.

podbuilder/podbuilder.go
```diff
@@ -310,6 +310,14 @@ func (pb *PodBuilder) newPod() *corev1.Pod {
 		})
 	}
 
+	pb.run.Env = append(pb.run.Env, corev1.EnvVar{
+		Name:  "POD_TIMEOUT",
```
codeblooded, line 314: nit: The name of this environment variable should be a global, like config.PodTimeoutEnv, to match the rest of the environment variables.

codeblooded (approved):
A couple comments to consider, but I will approve b/c it overall lgtm.

config/defaults.go
wanlin31: I agree it is bit ambiguous to not able to have a unit, but I think using two different names for the same variable is also a bit confusing. I can add more comments inside the code to improve readability of the code.

config/defaults.go
wanlin31: I think we called defs.Validate() to run the check at the configure time. I think I did not understand your question fully.

ybbbby (approved):

PR #178, opened 2021-06-08
Style fixes

codeblooded:
This PR fixes some inconsistent styling, license and documentation issues.

### DIFF
--- a/clientset/client.go
+++ b/clientset/client.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2021 gRPC authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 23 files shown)


paulosjca (changes requested):
Files that were created in 2020 should not have a copyright date of 2021. I randomly checked kubehelpers/pods.go, but this applies to all.

codeblooded:
@paulosjca I can move it back, but copyright dates are allowed to be bumped during changes at the owner's discretion. This moves the date of release to public domain forward, preserving it as a copyrighted work for a longer period of time. At the same time, the prior commit would retain the previous date. For example, see the bottom of the sidebar on https://youtube.com. It has been bumped to `© 2021 Google LLC`. I would not say YouTube has become a substantial new work between 2020 and 2021. Would you still like me to revert?

paulosjca (approved):

PR #180, opened 2021-06-10
Change the way ready containers find worker pods

ybbbby:
Now, the ready container would accept a command line argument to specify a load test name. It would find all worker pod (not driver) owned by the load test and record their ip addresses.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ uninstall: uninstall-crd uninstall-rbac
 
 # Install CRDs into a cluster
 install-crd: manifests
-	kustomize build config/crd | kubectl apply -f - 
+	kustomize build config/crd | kubectl apply -f -
 
 # Uninstall CRDs from a cluster
 uninstall-crd: manifests
--- a/Makefile
+++ b/Makefile
@@ -65,26 +65,26 @@ delete_prebuilt_workers: $(TOOLSPREREQ)
 	go build $(GOARGS) -o bin/delete_prebuilt_workers tools/delete_prebuilt_workers/delete_prebuilt_workers.go
 
 # Install both CRDs and RBACs
-install: installCRD installRBAC
+install: install-crd install-rbac
 
 # Uninstall both CRDs and RBACs
-uninstall: uninstallCRD uninstallRBAC
+uninstall: uninstall-crd uninstall-rbac
 
 # Install CRDs into a cluster
-installCRD: manifests
+install-crd: manifests
 	kustomize build config/crd | kubectl apply -f - 
 
 # Uninstall CRDs from a cluster
-uninstallCRD: manifests
+uninstall-crd: manifests
 	kustomize build config/crd | kubectl delete -f -
 
 
 # Install RBACs into a cluster
-installRBAC: manifests
+install-rbac: manifests
 	kustomize build config/rbac | kubectl apply -f -
 
 # Uninstall RBACs from a cluster
-uninstallRBAC: manifests
+uninstall-rbac: manifests
 	kustomize build config/rbac | kubectl delete -f -
 
 # Deploy both controller and cleanup_agent to the cluster
--- a/Makefile
+++ b/Makefile
@@ -65,12 +65,21 @@ delete_prebuilt_workers: $(TOOLSPREREQ)
 	go build $(GOARGS) -o bin/delete_prebuilt_workers tools/delete_prebuilt_workers/delete_prebuilt_workers.go
 
 # Install CRDs into a cluster
-install: manifests
-	kustomize build config/crd | kubectl apply -f -
+installCRD: manifests
+	kustomize build config/crd | kubectl apply -f - 
 
 # Uninstall CRDs from a cluster
-uninstall: manifests
+uninstallCRD: manifests
 	kustomize build config/crd | kubectl delete -f -
+
+
+# Install RBACs into a cluster
+installRBAC: manifests
+	kustomize build config/rbac | kubectl apply -f -
+
+# Uninstall RBACs from a cluster
+uninstallRBAC: manifests
+	kustomize build config/rbac | kubectl delete -f -
 
 # Deploy both controller and cleanup_agent to the cluster
 deploy: deploy-controller deploy-cleanup-agent
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,12 @@ prepare_prebuilt_workers: $(TOOLSPREREQ)
 delete_prebuilt_workers: $(TOOLSPREREQ)
 	go build $(GOARGS) -o bin/delete_prebuilt_workers tools/delete_prebuilt_workers/delete_prebuilt_workers.go
 
+# Install both CRDs and RBACs
+install: installCRD installRBAC
+
+# Uninstall both CRDs and RBACs
+uninstall: uninstallCRD uninstallRBAC
+
 # Install CRDs into a cluster
 installCRD: manifests
 	kustomize build config/crd | kubectl apply -f - 
--- a/containers/init/ready/README.md
+++ b/containers/init/ready/README.md
@@ -1,13 +1,13 @@
 # Ready
 
-Ready is a container that waits for a list of pods within a loadtest to
+Ready is a container that waits for a list of pods within a load test to
 become available. It exits successfully when all worker pods are ready, writing a
 comma-separated list of their IP addresses to a file. It exits unsuccessfully if
 a timeout was exceeded before all pods were ready.
 
 ## Usage
 
-The container relies on command line argument to specify the loadtest's name.
+The container relies on command line argument to specify the load test's name.
 For example,
 
 ```shell
--- a/containers/init/ready/README.md
+++ b/containers/init/ready/README.md
@@ -8,13 +8,12 @@ a timeout was exceeded before all pods were ready.
 ## Usage
 
 The container relies on command line argument to specify the load test's name.
-For example,
+For example,, the following waits for all worker pods which belong to the LoadTest
+named `<LOADTEST_NAME>`:
 
 ```shell
-go run ready.go LOADTEST_NAME
+go run ready.go <LOADTEST_NAME>
 ```
-
-Will wait for all worker pods which belongs to LOADTEST_NAME
 
 Meanwhile, users can set environment variables to override some defaults:
 
--- a/containers/init/ready/README.md
+++ b/containers/init/ready/README.md
@@ -1,25 +1,20 @@
 # Ready
 
-Ready is a container that waits for a list of pods with specific labels to
-become available. It exits successfully when all pods are ready, writing a
+Ready is a container that waits for a list of pods within a loadtest to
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -65,13 +65,22 @@ delete_prebuilt_workers: $(TOOLSPREREQ)
 	go build $(GOARGS) -o bin/delete_prebuilt_workers tools/delete_prebuilt_workers/delete_prebuilt_workers.go
 
 # Install CRDs into a cluster
-install: manifests
-	kustomize build config/crd | kubectl apply -f -
+installCRD: manifests
```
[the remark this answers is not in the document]
ybbbby: Done

containers/init/ready/README.md
```diff
@@ -1,25 +1,20 @@
 # Ready
 
-Ready is a container that waits for a list of pods with specific labels to
-become available. It exits successfully when all pods are ready, writing a
+Ready is a container that waits for a list of pods within a loadtest to
```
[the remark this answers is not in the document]
ybbbby: Done

paulosjca (approved):
Nit: Please update the PR title to be more descriptive before submitting.

paulosjca (approved):

containers/init/ready/ready.go
```diff
@@ -216,20 +211,46 @@ func main() {
...
+	})
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		if err != rest.ErrNotInCluster {
+			log.Fatalf("failed to connect within cluster: %v", err)
+		}
 		if err != nil {
-			log.Fatalf("failed to read kubeconfig: %v", err)
+			log.Fatalf("could not find a home directory for user: %v", err)
```
wanlin31, line 230: nit: is this line executed for sure? The line 225 and line 229 are checking the same condition, right?

wanlin31, line 230: nit: is this line executed for sure? The line 225 and line 229 are checking the same condition, so you will not get to any line below 231. I might have missed something.

wanlin31 (approved):

containers/init/ready/README.md
```diff
@@ -1,25 +1,20 @@
...
-go run ready.go role=server,proxy=envoy role=client role=client
+go run ready.go LOADTEST_NAME
 ```
 
-Will wait for three pods:
-
-1. pod with labels role=server AND proxy=envoy
-2. pod with label role=client
-3. another pod with label role=client
+Will wait for all worker pods which belongs to LOADTEST_NAME
```
codeblooded, line 17: nit: since there is no longer a list, it may read better if we merge lines 11 and 17:

````suggestion
For example, the following waits for all worker pods which belong to the LoadTest named `<LOADTEST_NAME>`:

```shell
go run ready.go <LOADTEST_NAME>
```
````

Also, this doesn't have to be fixed in this PR (and I apparently broke the convention myself (sorry))... but generally command line arguments adopt the following conventions:

- Angle brackets, such as `<ARG-NAME>`, indicate the argument is required.
- Square brackets, such as `[ARG-NAME]`, indicate the argument is optional.

You will sometimes see no angle brackets, such as the [NetBSD style](https://en.wikipedia.org/wiki/Usage_message#Examples) or the [Google Developer documentation](https://developers.google.com/style/code-syntax). However, the Google developer documentation switches the color and font for required variables. gcloud uses the angle brackets on the command line. If we don't, it can be hard to determine what is a command and what is a required argument. Most of the tools we use rely on angle brackets. For example,

- `git -h`:
  ```
  usage: git [--version] [--help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--super-prefix=<path>] [--config-env=<name>=<envvar>]
           <command> [<args>]
  ```

- `gcloud compute project-info -h`:
  ```
   Usage: gcloud compute project-info [optional flags] <command>
       command may be             add-metadata | describe | remove-metadata |
                                  set-usage-bucket | update
  ```

The Python [docopt](http://docopt.org/) project provides some good conventions. They actually generate a command line argument and flag parser using your usage string documentation.

File: containers/init/ready/ready.go [hunk 5]
```diff
@@ -70,21 +75,9 @@ type PodLister interface {
...
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to parse selector #%d (%s): %v", i+1, arg, err)
-		}
-
-		selectors = append(selectors, selector)
-	}
-
-	return selectors, nil
+// LoadtestGetter fetch a load test with a specific name.
+type LoadtestGetter interface {
```
codeblooded, line 79: nit: Since the object is a `LoadTest`, this should be `LoadTestGetter`. 

```suggestion
type LoadTestGetter interface {
```

containers/init/ready/ready.go
```diff
@@ -125,83 +118,85 @@ func findDriverPort(pod *corev1.Pod) int32 {
...
-// WaitForReadyPods blocks until pods with matching label selectors are ready.
+// WaitForReadyPods blocks until all worker pods belonging to the load test are
+// ready.
 // It accepts a context, allowing a timeout or deadline to be specified. When
 // all pods are ready, it returns a slice of strings with the IP address and
-// driver port for each matching pod. The order will match the label selectors.
-//
-// The syntax for the selectors is defined in the Parse function documented at
-// pkg.go.dev/k8s.io/apimachinery/pkg/labels.
+// driver port for each matching pod. server pod would come before client pod.
```
codeblooded, line 125: nit: We can adjust the comment so it takes up the full span of 80 characters.

containers/init/ready/ready.go
```diff
@@ -125,83 +118,85 @@ func findDriverPort(pod *corev1.Pod) int32 {
...
-	for range selectors {
-		podAddresses = append(podAddresses, "")
-	}
-
-	matchCount := 0
+	var loadtest *grpcv1.LoadTest
+	var clientPodAddresses []string
+	var serverPodAddresses []string
+	clientMatchCount := 0
+	serverMatchCount := 0
```
codeblooded, line 144: nit: We should place a space between lines 139 and 140 since they serve as two logically distinct purposes. Or, we could move lines 140-144 to be before line 134. Placing `var` declarations at the top is usually a good thing. If the function is super long, it sometimes helps to place them closer to where they are used though. The only other time that they generally don't belong as the very first thing is when there is a conditional that would prevent them from ever being needed. In this situation, the conditional in lines 136-139 does not return... so there is no reason they cannot be at the top for clarity.

containers/init/ready/ready.go
```diff
@@ -125,83 +118,84 @@ func findDriverPort(pod *corev1.Pod) int32 {
...
-					driverPort := findDriverPort(&pod)
-					podAddresses[i] = fmt.Sprintf("%s:%d", ip, driverPort)
-					matchingPods[pod.Name] = true
-					matchCount++
-					break
-				}
+			matchingPods[pod.Name] = true
+			ip := pod.Status.PodIP
+			driverPort := findDriverPort(pod)
+			if pod.GetLabels()[config.RoleLabel] == config.ServerRole {
```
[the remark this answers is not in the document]
codeblooded: Yes, the order matters. The servers must always be listed before the clients. This is how the driver understands what is a server.

containers/init/ready/ready.go
```diff
@@ -125,83 +118,85 @@ func findDriverPort(pod *corev1.Pod) int32 {
...
 		}
-
-		for _, pod := range podList.Items {
-			if !isPodReady(&pod) {
+		ownedPods := status.PodsForLoadTest(loadtest, podList.Items)
+		for _, pod := range ownedPods {
+			if !isPodReady(pod) {
+				continue
+			}
+			if pod.GetLabels()[config.RoleLabel] == config.DriverRole {
```
codeblooded, line 175: We should not need the GetLabels method. The metadata piece of the pod uses composition, so we should just be able to do `pod.Labels[config.RoleLabel]`.

containers/init/ready/ready_test.go
```diff
@@ -98,56 +103,48 @@ var _ = Describe("WaitForReadyPods", func() {
...
+			Loadtest: newLoadTestWithMultipleClientsAndServers(2, 0),
 		}
 
-		_, err := WaitForReadyPods(ctx, mock, []string{"role=driver,loadtest=loadtest-1"})
+		_, err := WaitForReadyPods(ctx, loadGetterMock, podListerMock, "test name")
 		Expect(err).To(HaveOccurred())
 	})
 
-	It("timeout reached when not ready pod matches", func() {
+	It("timeout reached when no ready pod matches", func() {
```
codeblooded, line 114: great catch 👍🏻

containers/init/ready/ready_test.go
```diff
@@ -245,3 +244,20 @@ func (plm *PodListerMock) List(opts metav1.ListOptions) (*corev1.PodList, error)
...
+
+func (lgm *LoadtestGetterMock) Get(testName string, opts metav1.GetOptions) (*grpcv1.LoadTest, error) {
+	time.Sleep(lgm.SleepDuration)
+
+	if lgm.Error != nil {
+		return nil, lgm.Error
+	}
+
+	return lgm.Loadtest, nil
+}
```
codeblooded, line 263: Great job! I would only change the name to `LoadTestGetterMock` to match the interface. You are doing a great job with our tests 😄

codeblooded (approved):
A few nits to consider but overall great job @ybbbby! LGTM. I'll mark as approved and you can choose to address them here before merging or in a follow-up PR.

containers/init/ready/README.md
ybbbby: Thanks for providing extra information! Very interesting :)

containers/init/ready/ready.go
ybbbby: Done

containers/init/ready/ready.go
ybbbby: Done

containers/init/ready/ready.go
ybbbby: Done, appreciate it!

containers/init/ready/ready.go
ybbbby: Done

containers/init/ready/ready_test.go
ybbbby: Done, thanks:)

containers/init/ready/README.md
paulosjca: FYI, conventions that are good for usage strings may not be good for markdown documentation. It is not important here because we can always fix it later, so let's merge the change and work on a release.

containers/init/ready/ready.go
paulosjca: Placing variables close to where they are used is generally a good thing too, so I wouldn't say that there is a preference for placing them at the top.

PR #182, opened 2021-06-12
Add timeout for workers

wanlin31:
This commit adds timeout for workers. Due to the complication of
directly pass the command and argument through LoadTest
configuration yaml files, a script is need to start the worker
with timeout. To make the script work, the argument
--driver_port of the run container has to be exposed. Currently
--driver_port is set as an environmental variable and passed to
worker at starting time.


wanlin31:
This is just a demonstration of the approach I currently think of, requesting a review just to make sure that the idea is okay.

wanlin31:
After the $DRIVER_PORT is exposed, there are no additional argument appended through controller. To further simplify the change, instead of creating a wrapper script for each worker the command was fed to bash as a string command. This ensure the same level of flexibly.

I have test cxx and go so far, both work as expected. Will continuous to test the rest of the tests.

wanlin31:
All tests (regular and prebuilt) are verified through running a controller against the benchmark-dev1 cluster. Will try all test with a deployed controller, but should be okay, will update further. Open pr for review of the ideas.

config/samples/csharp_example_loadtest.yaml
```diff
@@ -24,10 +24,12 @@ spec:
     name: '0'
     run:
       args:
-      - exec
-      - qps_worker/Grpc.IntegrationTesting.QpsWorker.dll
+      - -c
+      - "timeout --kill-after=$KILL_AFTER $POD_TIMEOUT dotnet exec \
```
paulosjca, line 28: I would try to make these changes in the templates in grpc/grpc, and then generate the configuration from there.

I have two expectations for what would happen:

1. The "timeout" string will be formatted as a multiline string, with |, which probably has better readability.
2. There may be issues with the dollar signs in this string, which may need to be escaped.
3. Assuming that (1) is true, we may want to format it as "${KILL_AFTER}" and so on (perhaps also escaping the dollar sign with $$).

containers/runtime/php7/run_php7_worker.sh
```diff
@@ -13,6 +13,25 @@
...
+
+echo "Server port: $input_server_port"
+
 source /usr/local/rvm/scripts/rvm
 export GEM_HOME=/src/workspace/saved/bundle/
-ruby src/ruby/qps/proxy-worker.rb
+
+if [ -z "$input_server_port" ]; then
+  echo "Server port is not set, starting the worker without server port provided"
+  timeout --kill-after=$KILL_AFTER $POD_TIMEOUT ruby src/ruby/qps/proxy-worker.rb \
```
paulosjca, line 30: I would make sure to put "${KILL_AFTER}" everywhere and so on.

You can run "shellcheck" on your shell script to see what it finds (there is an extension for that in vscode too, not sure about other editors).

podbuilder/podbuilder.go
```diff
@@ -152,7 +152,10 @@ func (pb *PodBuilder) PodForClient(client *grpcv1.Client) (*corev1.Pod, error) {
```
ybbbby, line 152: Is it possible that we add "timeout --kill-after=$KILL_AFTER $POD_TIMEOUT"  at line 189 when we create pod instead in each yaml file?

config/samples/csharp_example_loadtest.yaml
wanlin31: I will try that and running a CI from my fork (grpc_e2e_performance_v2.sh not the gke one), see if anything wired happened.

ybbbby (commented):
Nice job! Wanlin

config/samples/php7_example_loadtest.yaml
```diff
@@ -25,7 +25,7 @@ spec:
     run:
       command:
       - bash
-      - /run_scripts/run_worker.sh
+      - /run_scripts/run_php7_worker.sh
```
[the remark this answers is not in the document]
wanlin31: The new script is called "run_php7_worker", I changed it since I was going to add run_workers.sh to wrap all workers. It turn out to be that I don't have to. I can modify it back to run_worker.sh. :)

podbuilder/podbuilder.go
wanlin31: It is a great point, we would have been okay if we have run command for all languages exposed directly, but we have php7, which has to be wrapped with other reason. We might could have "timeout --kill-after=$KILL_AFTER $POD_TIMEOUT"  as a prefix, but for php7 it would be timeout the run_php7_worker.sh instead of the worker process itself.  For php7 we actually could add the timeout inside the wrapper , but that way php7 would have double timeout, not consistent with the rest of the languages. 

Generally, I think adding ENV in podbuilder is much safer since you can abandon it with no visible consequences, but if you add argument or command (or modified the command in anyway), it will always trying to execute it, if in the future we have to wrap more workers, this could be problematic.

@codeblooded @paulosjca

config/samples/csharp_example_loadtest.yaml
```diff
@@ -24,10 +24,13 @@ spec:
     name: '0'
     run:
       args:
-      - exec
-      - qps_worker/Grpc.IntegrationTesting.QpsWorker.dll
+      - -c
+      - |
+        timeout --kill-after=$${KILL_AFTER} $${POD_TIMEOUT} dotnet exec
+        qps_worker/Grpc.IntegrationTesting.QpsWorker.dll
+        --driver_port=$${DRIVER_PORT}
```
jtattermusch, line 31: qq: why the double `$$`? (I might be missing something).

config/samples/csharp_example_loadtest.yaml
```diff
@@ -24,10 +24,13 @@ spec:
     name: '0'
     run:
       args:
-      - exec
-      - qps_worker/Grpc.IntegrationTesting.QpsWorker.dll
+      - -c
+      - |
+        timeout --kill-after=$${KILL_AFTER} $${POD_TIMEOUT} dotnet exec
```
jtattermusch, line 29: improvement opportunity: I think just modifying the command to  run worker wrapped with timeout invocation works, but the downside is that all users need to remember to also invoke `timeout --kill-after=$${KILL_AFTER} $${POD_TIMEOUT}` in any custom experiments they run (e.g. experimenting with new images, languages, etc. where you need to specify the run worker command), otherwise they are risking their experiment will create pod don't terminate correctly.

So it would be nice if there was a way for the framework to automatically inject the `timeout --kill-after=$${KILL_AFTER} $${POD_TIMEOUT}` wrapper, without making that part of the worker run command explicitly.

Idea: docker images have the concept of `ENTRYPOINT` (which is simply a way to prepend any commands executed by the docker image with a command. https://docs.docker.com/engine/reference/builder/#entrypoint

so perhaps our Dockerimages can specify
```
ENTRYPOINT ["timeout", "--kill-after=${KILL_AFTER}", "${POD_TIMEOUT}"]
``` 

which would then inject the timeout logic automatically, without needing to make the run worker commands ugly
(and also needing to use the "bash -c" wrapping with confusing `$$`).
note that I haven't tried this, but it seems like something that would work.

config/samples/csharp_example_loadtest.yaml
wanlin31: Loadtest_template.py uses these examples to generate loadtest template, see [grpc/tools/run_tests/performance/templates/loadtest_template_basic_all_languages.yaml](https://github.com/grpc/grpc/blob/master/tools/run_tests/performance/templates/loadtest_template_basic_all_languages.yaml) and add some injection field, when the Loadtest_config.py takes the generated template to produce runnable loadtest configurations, it will require all fields with $ to be substitute with a actual values such as  ${client_pool}. KILL_AFTER, POD_TIMEOUT and DRIVER_PORT are used inside the container, we do not want to substitute actual values for them when we generating the configuration file, $$ is to escape. There could be other ways of doing it, open to modify if there are concerns.

config/samples/csharp_example_loadtest.yaml
wanlin31: I think it could work. We could also try to add `"timeout", "--kill-after=${KILL_AFTER}", "${POD_TIMEOUT}"` as a prefix  for the command when we created the pods from controller. Both of them would work fine when the start-worker-commands are exposed in the loadtest configuration file, but like php7, where we have to wrap the start-worker-commands with a script, both of our ways would put the timeout on the wrapper.

config/samples/csharp_example_loadtest.yaml
wanlin31: Moving forward we will not use $$, updating the pr.

config/samples/csharp_example_loadtest.yaml
wanlin31: We are moving forward on the ENTRYPOINT method you suggested, let me verify and update the pr

wanlin31: I tested with go example, seems like providing command through LoadTest configuration file override the ENTRYPOINT. (note that "The command field corresponds to entrypoint.")
"The command and arguments that you define in the configuration file override the default command and arguments provided by the container image. If you define args, but do not define a command, the default command is used with your new arguments."  https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/

containers/pre_built_workers/php7/run_worker.sh
```diff
@@ -13,6 +13,25 @@
 # limitations under the License.
 
 set -ex
+
```
paulosjca, line 16: In a follow-up PR, perhaps you can run  "shellcheck" on all shell scripts and fix outstanding issues. For now,  I am approving it as-is.

paulosjca (approved):

podbuilder/podbuilder.go
codeblooded: I prefer environment variables as mentioned by @wanlin31. Concatenating the strings in podbuilder seems fragile to me. It also may break or limit some commands.

config/samples/csharp_example_loadtest.yaml
jtattermusch: Ok, thanks for investigating this. What we have now seems good enough, so I'm going to approve.
We should keep our eyes open for possible simplifications in the future though. The current version is not as easy to maintain as I'd wish.

jtattermusch (approved):

config/samples/go_example_loadtest.yaml
```diff
@@ -27,9 +27,14 @@ spec:
     language: go
     name: '0'
     run:
+      args:
+      - -c
+      - |
+        timeout --kill-after="$${KILL_AFTER}" "$${POD_TIMEOUT}" \
```
jtattermusch, line 33: in this file, I can still see `$$` (they're gone in the other files).

wanlin31: I was experimenting with go, fix it, thank you for pointing it out!

containers/pre_built_workers/php7/run_worker.sh
wanlin31: Thank you for reminding me this, I was not able to use shellcheck earlier, now I have figured out how use use it now! The only thing is "source /usr/local/rvm/scripts/rvm" is not a local directory, shellcheck shows it as invalid path, how should I fix this?

podbuilder/podbuilder.go
wanlin31: And it probably need more work than I thought ( tried could not make it work easily...

PR #186, opened 2021-06-14
Update defaults_template.yaml for php7

wanlin31:


config/defaults_template.yaml
```diff
@@ -36,7 +36,7 @@ languages:
 
 - language: php7
   buildImage: "{{ .BuildImagePrefix }}php:{{ .Version }}"
```
paulosjca, line 38: Is the build image really "php"?

wanlin31: I think in the makefile we pushed them to php7 instead of php 
https://github.com/grpc/test-infra/blob/29ec42a3716e76ff3822aef64f0c356f16a6a787/Makefile#L224

paulosjca: > I think in the makefile we pushed them to php7 instead of php
> https://github.com/grpc/test-infra/blob/29ec42a3716e76ff3822aef64f0c356f16a6a787/Makefile#L224

Right, let's fix that too then. You can check in the latest release images.

wanlin31: I did not fully understand what else needed to be updated, I believe this is the only place left as php. The make targets are all updated to php7. Let me know if I missed something. :)

wanlin31: Right, sorry I missed it, I thought I changed it.

paulosjca (approved):

PR #188, opened 2021-06-17
Make run field optional in driver configuration.

wanlin31:
This commit changes grpcv1.Driver.Run from a required value to
an option value so that the driver's name could be set within
LoadTest configuration yaml file.

### DIFF
--- a/api/v1/loadtest_types.go
+++ b/api/v1/loadtest_types.go
@@ -161,6 +161,7 @@ type Driver struct {
 
 	// Run describes the run container, which is the runtime of the driver for
 	// the actual test.
+	// +optional
 	Run Run `json:"run"`
 }
 
--- a/config/samples/go_example_loadtest.yaml
+++ b/config/samples/go_example_loadtest.yaml
@@ -29,6 +29,9 @@ spec:
     run:
       command:
       - /src/workspace/bin/worker
+  driver:
+    language: cxx
+    name: '0'
   scenariosJSON: |
     {
       "scenarios": {
--- a/config/crd/bases/e2etest.grpc.io_loadtests.yaml
+++ b/config/crd/bases/e2etest.grpc.io_loadtests.yaml
@@ -802,7 +802,6 @@ spec:
                   type: object
               required:
               - language
-              - run
               type: object
             results:
               description: Results configures where the results of the test should


wanlin31:
Since grpcv1.Driver.Run was a required filed whenever we have the Run section in the LoadTest configuration yaml files, we could not only supply driver names. Two options:
(1)Shows in this PR: change grpcv1.Driver.Run to not be a required field, I think this way configure driver's information is much easier.
(2)Similar to [grpc/tools/run_tests/performance/templates/loadtest_template_prebuilt_all_languages.yaml](https://github.com/grpc/grpc/blob/master/tools/run_tests/performance/templates/loadtest_template_prebuilt_all_languages.yaml), have all required filed in the LoadTest configuration yaml files.

The Driver sections as a whole was an optional field we have never really rely on this run section for driver. Is there any concern of making grpcv1.Driver.Run and optional field?

config/samples/go_example_loadtest.yaml
```diff
@@ -29,6 +29,9 @@ spec:
     run:
       command:
       - /src/workspace/bin/worker
+  driver:
```
paulosjca, line 32: You should probably not change this as part of this PR, but only after the next release is deployed.

api/v1/loadtest_types.go
```diff
@@ -161,6 +161,7 @@ type Driver struct {
 
 	// Run describes the run container, which is the runtime of the driver for
 	// the actual test.
+	// +optional
 	Run Run `json:"run"`
```
codeblooded, line 165: Hold up a moment. This is what I meant in the meeting last week. Run cannot be optional, b/c you must always have a container to run.... at least an image. We always need a container to run.

In addition, this change violates the Kubernetes API conventions... we cannot make a non-pointer field `//+optional`. And optionals should have a `json:"run,omitempty"` tag. So this becomes `*Run` which is the fix we don't want and will break things later down the pipeline.

We need to discuss alternative ways to address this issue.

codeblooded (changes requested):

api/v1/loadtest_types.go
paulosjca: > We need to discuss alternative ways to address this issue.

The run field is also required in the client and server, but that doesn't require us to include an image. The alternative solution is to make the driver behave in the same way as the clients and servers, and I agree that it does not require making the run field optional.

paulosjca: It looks like "image" is optional. So the question may be what is the difference between the handling of image in these two places:

https://github.com/grpc/test-infra/blob/de9cdd98b55cb95db97b13356d2cacf10dbb1bf2/config/defaults.go#L195

https://github.com/grpc/test-infra/blob/de9cdd98b55cb95db97b13356d2cacf10dbb1bf2/config/defaults.go#L216

paulosjca: One issue with the above snippets is that they are picking driver and client images from the same image map, and that cannot be right.

paulosjca: Here is something interesting: if I make an assumption about the 'command' to run the driver, I can specify the name without specifying the image, and the test succeeds:

```yaml
  driver:
    language: cxx
    name: '0'
    run:
      command:
      - /src/driver/run.sh
```

paulosjca: It looks like it is possible to make it work with no assumptions, by always adding `driver`, `run` and `env` fields if they don't exist, resulting in the following: 

```yaml
  driver:
    language: cxx
    name: '0'
    run:
      env: []
```
I will update the config generator and the examples.

We still need to fix the logic that might apply worker images to the driver.

paulosjca: FYI, the final solution is in #189:

```yaml
  driver:
    language: cxx
    name: '0'
    run: {}
```

wanlin31: Thank you for the discussion. Currently in Driver struct, Run is a required filed, but every field inside Run struct itself is optional. So we could get aways with adding `run: {}` as mentioned by @paulosjca . @codeblooded Thank you for helping me learning the [Kubernetes's conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required).
I closed this PR since it is better addressed in #189.

PR #192, opened 2021-06-28
Update driver timeout

wanlin31:
This commit update where the timeout is applied for driver
process. The timeout was wrapped around the driver
process only, which left other process such as uploading 
results to BigQuery unguarded. The timeout is updated
to wrap around the whole process to make sure driver
pod will timeout after given time.

### DIFF
--- a/containers/runtime/driver/Dockerfile
+++ b/containers/runtime/driver/Dockerfile
@@ -87,4 +87,4 @@ ENV QPS_WORKERS_FILE=""
 ENV SCENARIOS_FILE="/src/driver/example.json"
 ENV BQ_RESULT_TABLE=""
 
-CMD ["bash", "-c", "timeout --kill-after=$KILL_AFTER $POD_TIMEOUT /src/driver/run.sh"]
+CMD ["bash", "-c", "timeout --kill-after=${KILL_AFTER} ${POD_TIMEOUT} /src/driver/run.sh"]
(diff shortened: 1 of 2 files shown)


wanlin31:
Timeout could not access function defined inside the script, we can:
(1) Timeout run.sh all together, as @ybbbby has figured out that even the timeout is on the wrapper script, the subprocess would be killed too.
(2) Create another script and timeout that script inside run.sh. 

I went with option 1 since it is simpler.

Reference: https://stackoverflow.com/questions/687948/timeout-a-command-in-bash-without-unnecessary-delay

containers/runtime/driver/Dockerfile
```diff
@@ -87,4 +87,4 @@ ENV QPS_WORKERS_FILE=""
 ENV SCENARIOS_FILE="/src/driver/example.json"
 ENV BQ_RESULT_TABLE=""
 
-CMD ["/src/driver/run.sh"]
+CMD ["bash", "-c", "timeout --kill-after=$KILL_AFTER $POD_TIMEOUT /src/driver/run.sh"]
```
paulosjca, line 90: Would it work if you would separate every argument here, "bash", "-c", "timeout", "--kill-after=${KILL_AFTER}" and so on? I would even like to see if we can have "" around the variables, but that may be too much.

wanlin31: I will try that

wanlin31: I have tried to use:
```
CMD ["bash", "-c", "timeout", "--kill-after=${KILL_AFTER}", "${POD_TIMEOUT}", " /src/driver/run.sh"]
```
Error with:
```
Try 'timeout --help' for more information.
```
Seems like the argument for `bash -c` has to be in a single string. 

I put the `{}` for the variable, it doesn't allow ""  to wrap around the variables.

paulosjca (approved):

wanlin31:
Verified timeout does apply to the subprocess.

PR #195, opened 2021-06-29
Feature/add php7_protobuf_c

wanlin31:
This commit add support for php7_protobuf_c tests. Also modify 
where the timeout is applied for php7 tests. Previously, the timeout
for php7 workers was wrapped directly on the worker binary, which
make it inconsistent with other workers. This Pr also fix this issue.


config/samples/php7_protobuf_c_example_loadtest.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - -c
+      - |
+        timeout --kill-after="${KILL_AFTER}" "${POD_TIMEOUT}" \
+        /run_scripts/run_protobuf_c_worker.sh
+      command:
+      - bash
+  driver:
+    language: cxx
+    name: '0'
+    run: { }
```
paulosjca, line 36: I don't think this should have a space. Was it generated from grpc/grpc? In any case, it should have the same format as the others.

wanlin31: Fixed :)

paulosjca (approved):

ybbbby (approved):
LGTM!

PR #198, opened 2021-07-12
Initial fixes for xml report generation.

paulosjca:
* Count errors instead of failures.
* Stop counting warnings as errors.
* Keep `id` attribute but only for `testsuite`, as a count starting from zero.
* Rename package to `xunit`.

### DIFF
--- a/tools/cmd/runner/main.go
+++ b/tools/cmd/runner/main.go
@@ -23,7 +23,7 @@ import (
 	"time"
 
 	"github.com/grpc/test-infra/tools/runner"
-	"github.com/grpc/test-infra/tools/runner/junit"
+	"github.com/grpc/test-infra/tools/runner/xunit"
 )
 
 func main() {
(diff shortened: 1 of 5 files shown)


wanlin31 (approved):
LGTM. I have a quick question regarding ID, is it the same with the LoadTest's name?

paulosjca:
> LGTM. I have a quick question regarding ID, is it the same with the LoadTest's name?

Strictly speaking, this field is not necessarily, because it is not part of the format that is understood by kokoro. The only reason to include it at all is because it is part of the Junit schema:

https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd

What we were doing before is add `id` to `testsuites`, `testsuite` and `testcase`, and to fill it with something derived from the element's name. However, if you look at the Junit schema, it includes `id` only for `testsuite`, where it is a counter (`0`, `1`, `2`,...). This PR changes the `id` attribute to follow the Junit schema. In the future we may consider whether we should include it or not.

PR #199, opened 2021-07-14
Make ready container write out the annotation as JSON

wanlin31:


containers/init/ready/ready.go
```diff
@@ -52,6 +53,14 @@ const DefaultTimeout = 25 * time.Minute
 // unset, DefaultOutputFile will be used as the default.
 const OutputFileEnv = "READY_OUTPUT_FILE"
 
+// OutputMetadataEnv is the optional name of the file where the executable should
+// write a comma-separated list of IP addresses.
```
paulosjca, line 57: I think you mean "write the test metadata"?

containers/init/ready/ready.go
```diff
@@ -269,4 +278,19 @@ func main() {
...
+	}
+
+	outputMetadataFile := DefaultMetadataOutputFile
+	outputMetadataFileOverride, ok := os.LookupEnv(OutputMetadataEnv)
+	if ok {
+		outputMetadataFile = outputMetadataFileOverride
+	}
+
+	metaDataSet := test.ObjectMeta
+	metaDataBody, _ := json.Marshal(metaDataSet)
```
paulosjca, line 294: Would it make sense to check for an error here?

paulosjca (commented):
Generally, LGTM.

containers/init/ready/ready.go
wanlin31: Right, I was trying to copy the way we explained the constants, wanted to be consistent, forgot to change it. :) Thank you

containers/init/ready/ready.go
wanlin31: Right, I will add a check here.

Makefile
```diff
@@ -135,8 +135,12 @@ php7-image: ## Build the PHP7 test runtime container image.
 python-image: ## Build the Python test runtime container image.
 	docker build -t ${IMAGE_PREFIX}python:${TEST_INFRA_VERSION} containers/runtime/python
 
-ready-image: ## Build the ready init container image.
-	docker build -t ${INIT_IMAGE_PREFIX}ready:${TEST_INFRA_VERSION} -f containers/init/ready/Dockerfile .
+# Build the driver container image at the $DRIVER_VERSION
```
wanlin31, line 138: I am not sure if we should have a cache breaker here.
Pro: Make sure we definitely build our driver from a fresh repo
Con: Since we do not do this very often, not sure if this is necessary

ybbbby (commented):
LGTM, some NITs

ybbbby (approved):

containers/init/ready/ready.go
```diff
@@ -269,4 +278,22 @@ func main() {
...
+		log.Fatalf("failed to fetch loadtest for obtaining metadata: %v", err)
+	}
+
+	outputMetadataFile := DefaultMetadataOutputFile
+	outputMetadataFileOverride, ok := os.LookupEnv(OutputMetadataEnv)
+	if ok {
+		outputMetadataFile = outputMetadataFileOverride
+	}
+
+	metaDataSet := test.ObjectMeta
```
paulosjca, line 293: Is this the right information, i.e., only the "metadata"  field?

wanlin31: The information typically included are:
```
{
    "annotations": {
        "buildUrl": "wanlin-test",
        "gitActualCommit": "wanlin-test",
        "gitCommit": "wanlin-test",
        "jobName": "wanlin-test",
        "scenario": "go_generic_sync_streaming_ping_pong_secure",
        "uniquifier": "basic"
    },
    "creationTimestamp": "2021-08-10T22:06:37Z",
    "generation": 2,
    "labels": {
        "language": "go",
        "prefix": "examples"
    },
    "name": "examples-go-generic-sync-streaming-ping-pong-secure-basic",
    "namespace": "default",
    "resourceVersion": "139781494",
    "uid": "77870d26-8382-495e-b311-64b5e64dce40"
}
```
I ignored the annotations attached by kubectl, which only exit when we apply the LoadTest configuration through command line.
I have also omit the managedFields, which are not metadata.

ybbbby (approved):

containers/runtime/driver/run.sh
```diff
@@ -24,7 +24,9 @@ fi
 
 /src/code/bazel-bin/test/cpp/qps/qps_json_driver --quit=true
 
-if [ "$BQ_RESULT_TABLE" != "" ]
-then
+if [ "$BQ_RESULT_TABLE" != "" ]; then
```
paulosjca, line 27: This is fine. You could also use `-a` to put both conditions in one test.

paulosjca (approved):

containers/runtime/driver/run.sh
wanlin31: I have updated the two conditions to be in the same statement. Thank you.

PR #201, opened 2021-07-19
Ignore empty default pool

ybbbby:


controllers/loadtest_controller.go
```diff
@@ -287,6 +287,9 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		adjustAvailabilityForDefaults(status.DefaultServerPool, defaultServerPool)
 
 		for pool, requiredNodeCount := range missingPods.NodeCountByPool {
+			if requiredNodeCount == 0 {
```
paulosjca, line 290: I think this should be changed in the logic of `adjustAvailabilityForDefaults` instead. If `c == 0`, then we don't make changes to the node count by pool. Otherwise, if `defaultPoolName` is empty, we return an error; if it is not empty, then we add the count to the default pool count.

ybbbby: Done

controllers/loadtest_controller.go
```diff
@@ -276,15 +276,36 @@ func (r *LoadTestReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			}
 		}
 
-		adjustAvailabilityForDefaults := func(defaultPoolKey, defaultPoolName string) {
+		adjustAvailabilityForDefaults := func(defaultPoolKey, defaultPoolName string) bool {
 			if c, ok := missingPods.NodeCountByPool[defaultPoolKey]; ok {
```
paulosjca, line 280: Nit: you may be able to test for `ok && c > 0` in one go.

I might also then test for `defaultPoolName == ""` since it has a `return` in it, and then do the other branch without the else.

paulosjca (approved):

wanlin31 (approved):

controllers/loadtest_controller.go
ybbbby: Done, definitely more elegant!

PR #202, opened 2021-07-28
Separate deployment of cleanup agent from main process

wanlin31:
This commit separate the deployment of cleanup agent from the 
main deployment process. This is the first step of deprecating
the cleanup agent.

The LoadTests which run longer than timeoutSeconds are still
marked as "TimeoutErrored".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ CONTROLLER_IMG ?= ${IMAGE_PREFIX}controller:${TEST_INFRA_VERSION}
 CLEAN_IMG ?= ${IMAGE_PREFIX}cleanup:${TEST_INFRA_VERSION}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
+# Flag to decide if we deploy the cleanup agent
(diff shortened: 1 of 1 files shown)


ybbbby (approved):

Makefile
```diff
@@ -87,8 +89,14 @@ install-rbac: manifests
 uninstall-rbac: manifests
 	kustomize build config/rbac | kubectl delete --ignore-not-found=true -f -
 
-# Deploy both controller and cleanup_agent to the cluster
-deploy: deploy-controller deploy-cleanup-agent
+# Deploy both controller and cleanup_agent or controller only based on
+# the environment variable
+deploy:
```
ybbbby, line 94: I am not familiar with Makefile. This first `deploy:` seems to be empty. Maybe we can delete it?

paulosjca (approved):

Makefile
paulosjca: You are right, this line should be deleted.

paulosjca: I would also leave a space after "deploy:" in the lines that follow.

wanlin31:
I have update the PR based on the comments and merged it.

PR #205, opened 2021-07-30
Solve flakiness in kb3 cluster

ybbbby:
Remove resource quotes restriction.
Grant controller with events related operations.
Resolve a cherry-pick bug left in https://github.com/grpc/test-infra/pull/203

### DIFF
--- a/controllers/loadtest_controller.go
+++ b/controllers/loadtest_controller.go
@@ -58,6 +58,7 @@ type LoadTestReconciler struct {
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=nodes/status,verbs=get
+// +kubebuilder:rbac:groups="",resources=events,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;create;update
 
 // Reconcile attempts to bring the current state of the load test into agreement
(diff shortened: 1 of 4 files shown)


controllers/loadtest_controller.go
```diff
@@ -270,12 +271,12 @@ func (r *LoadTestReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		adjustAvailabilityForDefaults := func(defaultPoolKey, defaultPoolName string) bool {
 			if c, ok := missingPods.NodeCountByPool[defaultPoolKey]; ok && c > 0 {
 				if defaultPoolName == "" {
-					log.Error(errNonexistentPool, "default pool is not defined or does not existed in the cluster", "requestedDefaultPool", defaultPoolKey)
+					logger.Error(errNonexistentPool, "default pool is not defined or does not existed in the cluster", "requestedDefaultPool", defaultPoolKey)
```
wanlin31, line 274: Wondering how would it look like before.

controllers/loadtest_controller.go
```diff
@@ -58,6 +58,7 @@ type LoadTestReconciler struct {
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=nodes/status,verbs=get
+// +kubebuilder:rbac:groups="",resources=events,verbs=get;list;watch;create;update;patch;delete
```
wanlin31, line 61: Do we need all of the verbs? `get;list;watch;create;update;patch;delete`

controllers/loadtest_controller.go
ybbbby: I did modify locally, but failed to commit in my last pr. It would give build error.

wanlin31 (approved):
LGTM just have some questions

controllers/loadtest_controller.go
ybbbby: Good point, only `create` is observed when controller crashed and restarted. 
I assume we can grant controller all the power to deal with events, no?

paulosjca (approved):

PR #211, opened 2021-08-05
Allow go version to be specified for infra build.

paulosjca:
Default build image in kokoro workers is too old (1.12).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,16 +29,11 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-GOVERSION=$(shell go version | cut -f3 -d' ' | cut -c3-)
+# Golang command for build
+GOCMD ?= go
 
-# Make tools build compatible with go 1.12
-ifeq (1.13,$(shell echo -e "1.13\n$(GOVERSION)" | sort -V | head -n1))
-GOARGS=-trimpath
-TOOLSPREREQ=fmt vet
-else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -29,16 +29,11 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-GOVERSION=$(shell go version | cut -f3 -d' ' | cut -c3-)
+# Golang command for build
+GOCMD ?= go
```
ybbbby, line 33: Shall we make the default vision to be higher than 1.12?

Makefile
```diff
@@ -29,16 +29,11 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-GOVERSION=$(shell go version | cut -f3 -d' ' | cut -c3-)
+# Golang command for build
```
ybbbby, line 32: Shall we add a comment specify that golang version has to be higher than 1.13?

ybbbby (approved):

Makefile
paulosjca: The whole project specifies go 1.16, so we are not officially supporting anything lower. With the upgrade to kubebuilder 3, it looks like the dependencies require at least go 1.15. The tools require only go 1.13, but I would rather leave it open and update to go 1.16.6 in https://github.com/grpc/grpc/pull/26881.

PR #215, opened 2021-08-19
save node info to a json file

ybbbby:
Driver's ready container would record workers' node info along the way of waiting them. The nodes information would be recorded in node_info.json, which would be saved in BigQuery.

### DIFF
--- a/containers/init/ready/ready.go
+++ b/containers/init/ready/ready.go
@@ -97,11 +97,15 @@ type PodLister interface {
 type LoadTestGetter interface {
 	Get(context.Context, string, metav1.GetOptions) (*grpcv1.LoadTest, error)
 }
+
+// NodeInfo contain a pod name, pod IP and node name for one worker or driver.
 type NodeInfo struct {
 	Name     string
 	PodIP    string
 	NodeName string
 }
+
+// NodesInfo contain all pods' NodeInfo included in a load test.
 type NodesInfo struct {
 	Driver  NodeInfo
 	Servers []NodeInfo
--- a/containers/init/ready/ready.go
+++ b/containers/init/ready/ready.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net"
 	"os"
 	"strings"
 	"time"
@@ -57,13 +58,21 @@ const OutputFileEnv = "READY_OUTPUT_FILE"
 // write all metadata.
 const OutputMetadataEnv = "METADATA_OUTPUT_FILE"
 
-// DefaultMetadataOutputFile is the name of the default file where the executable should
-// write the metadata.
-const DefaultMetadataOutputFile = "/tmp/metadata.json"
+// OutputNodeInfoEnv is the optional name of the file where the executable should
+// write the node information.
+const OutputNodeInfoEnv = "NODE_INFO_OUTPUT_FILE"
 
 // DefaultOutputFile is the name of the default file where the executable should
 // write the comma-separated list of IP addresses.
 const DefaultOutputFile = "/tmp/loadtest_workers"
+
+// DefaultMetadataOutputFile is the name of the default file where the executable should
+// write the metadata.
+const DefaultMetadataOutputFile = "/tmp/metadata.json"
+
+// DefaultNodeInfoOutputFile is the name of the default file where the executable should
+// write the node infomation.
+const DefaultNodeInfoOutputFile = "/tmp/node_info"
 
 // DefaultDriverPort is the default port for communication between the driver
 // and worker pods. When another port could not be found on a pod, this port is
@@ -87,6 +96,16 @@ type PodLister interface {
 // LoadTestGetter fetch a load test with a specific name.
 type LoadTestGetter interface {
 	Get(context.Context, string, metav1.GetOptions) (*grpcv1.LoadTest, error)
+}
+type NodeInfo struct {
+	Name     string
+	PodIP    string
+	NodeName string
+}
+type NodesInfo struct {
+	Driver  NodeInfo
+	Servers []NodeInfo
+	Clients []NodeInfo
 }
 
 // isPodReady returns true if the pod has been assigned an IP address and all of
@@ -138,12 +157,14 @@ func findDriverPort(pod *corev1.Pod) int32 {
 //
 // If the timeout is exceeded or there is a problem communicating with the
 // Kubernetes API, an error is returned.
-func WaitForReadyPods(ctx context.Context, ltg LoadTestGetter, pl PodLister, testName string) ([]string, error) {
+func WaitForReadyPods(ctx context.Context, ltg LoadTestGetter, pl PodLister, testName string) ([]string, *NodesInfo, error) {
 	var loadtest *grpcv1.LoadTest
 	var clientPodAddresses []string
 	var serverPodAddresses []string
+	var nodesInfo NodesInfo
 	clientMatchCount := 0
 	serverMatchCount := 0
+	driverMatched := false
 	timeoutsEnabled := true
 	matchingPods := make(map[string]bool)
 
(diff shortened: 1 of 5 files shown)


containers/init/ready/ready.go
```diff
@@ -89,6 +98,20 @@ type LoadTestGetter interface {
...
 }
 
+// NodeInfo contain pod name, pod IP and node name in which the pod reside for one worker or driver.
+type NodeInfo struct {
+	Name     string
+	PodIP    string
+	NodeName string
+}
+
+// NodesInfo contain all pods' NodeInfo included in a load test.
```
paulosjca, line 108: Perhaps "NodesInfo contains NodeInfo for all pods included in a load test."

containers/init/ready/ready_test.go
```diff
@@ -61,15 +62,24 @@ var _ = Describe("WaitForReadyPods", func() {
...
 			Loadtest: newLoadTestWithMultipleClientsAndServers(0, 0),
 		}
-		podAddresses, err := WaitForReadyPods(ctx, loadTestGetterMock, podListerMock, "test name")
+		podAddresses, nodesInfo, err := WaitForReadyPods(ctx, loadTestGetterMock, podListerMock, "test name")
 		Expect(err).ToNot(HaveOccurred())
 		Expect(podAddresses).To(BeEmpty())
+		Expect(*nodesInfo).To(Equal(NodesInfo{
+			Driver: NodeInfo{
+				Name:     driverPod.Name,
+				PodIP:    net.JoinHostPort(driverPod.Status.PodIP, "10000"),
```
paulosjca, line 79: In general, I think PodIP should be just the IP address, not a port.

containers/init/ready/ready.go
ybbbby: Done, thanks!

containers/init/ready/ready_test.go
ybbbby: Make sense, done.

containers/init/ready/ready.go
```diff
@@ -57,14 +58,22 @@ const OutputFileEnv = "READY_OUTPUT_FILE"
...
 // write the comma-separated list of IP addresses.
 const DefaultOutputFile = "/tmp/loadtest_workers"
 
+// DefaultMetadataOutputFile is the name of the default file where the executable should
+// write the metadata.
+const DefaultMetadataOutputFile = "/tmp/metadata.json"
+
+// DefaultNodeInfoOutputFile is the name of the default file where the executable should
+// write the node infomation.
+const DefaultNodeInfoOutputFile = "/tmp/node_info"
```
wanlin31, line 75: when you output the file is it automatically knows the format? I had it with .json, not sure if it is okay to not have it.

containers/init/ready/ready.go
```diff
@@ -57,14 +58,22 @@ const OutputFileEnv = "READY_OUTPUT_FILE"
 // write all metadata.
 const OutputMetadataEnv = "METADATA_OUTPUT_FILE"
 
-// DefaultMetadataOutputFile is the name of the default file where the executable should
-// write the metadata.
-const DefaultMetadataOutputFile = "/tmp/metadata.json"
+// OutputNodeInfoEnv is the optional name of the file where the executable should
```
wanlin31, line 61: Can I ask why is it optional?

containers/init/ready/ready.go
```diff
@@ -84,11 +93,25 @@ type PodLister interface {
 	List(context.Context, metav1.ListOptions) (*corev1.PodList, error)
 }
 
-// LoadTestGetter fetch a load test with a specific name.
+// LoadTestGetter fetches a load test with a specific name.
```
wanlin31, line 96: Nice catch!

containers/init/ready/ready.go
```diff
@@ -178,10 +203,18 @@ func WaitForReadyPods(ctx context.Context, ltg LoadTestGetter, pl PodLister, tes
...
-			if !isPodReady(pod) {
+			if pod.Labels[config.RoleLabel] == config.DriverRole {
+				if !driverMatched && pod.Status.PodIP != "" {
+					nodesInfo.Driver = NodeInfo{
+						Name:     pod.Name,
+						PodIP:    pod.Status.PodIP,
+						NodeName: pod.Spec.NodeName,
+					}
+					driverMatched = true
+				}
```
wanlin31, line 214: Why this portion is before checking if the pod is ready? Part of the isPodReady part is checking if the IP exist.

containers/init/ready/ready.go
```diff
@@ -296,4 +345,10 @@ func main() {
 		log.Fatalf("failed to marshal metaData for loadtest %s: %v", test.Name, err)
 	}
 	ioutil.WriteFile(outputMetadataFile, metaDataBody, 0777)
+
+	nodeInfoFileBody, err := json.Marshal(*nodesInfo)
+	if err != nil {
+		log.Fatalf("failed to marshal node information for loadtest %s: %v", test.Name, err)
```
wanlin31, line 351: nit: maybe nodes information?

wanlin31 (commented):
Generally looks good to me!

containers/init/ready/ready.go
ybbbby: Yeah, the format should be json. By adding .json, it would be more clear. Thanks!

containers/init/ready/ready.go
ybbbby: I assume those environment variables are all optional, like "OutputMetadataEnv, OutputFileEnv".

containers/init/ready/ready.go
ybbbby: Yeah, good question. since ready.go is running in driver's ready init container, itself would not be marked as ready. So in order to fetch driver's ip and node name, we have to grab it before ready check.

containers/init/ready/ready.go
ybbbby: Thanks!

containers/init/ready/ready.go
wanlin31: I think the variable itself may not be used but it is not really optional though. The name it sets for the file is the same with default name, it is a bit confusing  to say it is optional.

containers/runtime/driver/run.sh
```diff
@@ -24,7 +24,8 @@ fi
 
 /src/code/bazel-bin/test/cpp/qps/qps_json_driver --quit=true
 
-if [ -n "${BQ_RESULT_TABLE}" ] && [ -r "${METADATA_OUTPUT_FILE}" ]; then
+if [ -n "${BQ_RESULT_TABLE}" ] && [ -r "${METADATA_OUTPUT_FILE}" ] && [ -r "${NODE_INFO_OUTPUT_FILE}" ]; then
```
paulosjca, line 27: Reading more carefully, I don't think this logic is correct.

I think it should be something like this:

```shell
if [-n "${BQ_RESULT_TABLE}"]; then
  [ -r "${METADATA_OUTPUT_FILE}" ] && cp "${METADATA_OUTPUT_FILE}" metadata.json
  [ -r "${NODE_INFO_OUTPUT_FILE}" ] && cp "${NODE_INFO_OUTPUT_FILE}" node_info.json
  python3 /src/code/tools/run_tests/performance/bq_upload_result.py --bq_result_table="${BQ_RESULT_TABLE}"
fi
```

Unless we are assuming that the metadata and node info files must always be created, but even in that case I think the data should be saved to BigQuery even if the metadata or node_info files are missing.

containers/init/ready/ready.go
ybbbby: I am confused here. The comment for `OutputNodeInfoEnv` is rooted from `OutputMetadataEnv` and `OutputFileEnv`. Are you suggesting that we should change them too?

containers/runtime/driver/run.sh
ybbbby: done

paulosjca (approved):

containers/init/ready/ready.go
wanlin31: No, you are right, diff that github shows lead me think you modified the default to optional. I think it is fine to use optional that all others has been using. I just felt that these ENVs are actually hardcoded in podbuilder, they are the actual default (i.e if we don't do anything)

wanlin31 (approved):

PR #216, opened 2021-08-20
End xml reports in newline.

paulosjca:


ybbbby (approved):
It looks good to me. Where are those xml reports can be checked?

wanlin31 (approved):

paulosjca:
> It looks good to me. Where are those xml reports can be checked?

The reports are generated by the runner when the `-o` flag is set. To generate an example:

```shell
./bin/runner -annotation-key= -i config/samples/go_example_loadtest.yaml -c :2 -o sponge_log.xml
```

In test fusion, you can find the xml file by selecting the "runner" target and going to the "Artifacts" tab.

PR #80, opened 2020-10-27
Remove load test controller from top-level

codeblooded:
The grpc/test-infra repository contained the manager for the load test controller and its Dockerfile at the top-most level. This change moves it the image into the containers/ directory and introduces a cmd/ directory for the manager executable. It also renames the make target from "manager" to "controller".

PR #223, opened 2021-10-01
Get rid of rvm version warning

ybbbby:
I found that no .bashrc file edits needed, and it could get rid of the warning. Need confirmation from @wanlin31

### DIFF
--- a/containers/init/build/php7/Dockerfile
+++ b/containers/init/build/php7/Dockerfile
@@ -30,12 +30,10 @@ RUN gpg2 --keyserver keyserver.ubuntu.com --recv-keys 409B6B1796C275462A17031138
 RUN \curl -sSL https://get.rvm.io | bash -s stable
 
 # Install Ruby 2.5
-RUN apt-get update && apt-get install -y procps && apt-get clean
+RUN apt-get --allow-releaseinfo-change update && apt-get install -y procps && apt-get clean
 RUN /bin/bash -l -c "rvm install ruby-2.5"
 RUN /bin/bash -l -c "rvm use --default ruby-2.5"
 RUN /bin/bash -l -c "echo 'gem: --no-document' > ~/.gemrc"
-RUN /bin/bash -l -c "echo 'export PATH=/usr/local/rvm/bin:$PATH' >> ~/.bashrc"
-RUN /bin/bash -l -c "echo 'rvm --default use ruby-2.5' >> ~/.bashrc"
 RUN /bin/bash -l -c "gem install bundler --no-document -v 1.9"
 
 # Install composer
--- a/containers/pre_built_workers/php7/Dockerfile
+++ b/containers/pre_built_workers/php7/Dockerfile
@@ -31,12 +31,10 @@ RUN gpg2 --keyserver keyserver.ubuntu.com --recv-keys 409B6B1796C275462A17031138
 RUN \curl -sSL https://get.rvm.io | bash -s stable
 
 # Install Ruby 2.5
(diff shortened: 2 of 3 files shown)


paulosjca (approved):

containers/init/build/php7/Dockerfile
```diff
@@ -30,12 +30,10 @@ RUN gpg2 --keyserver keyserver.ubuntu.com --recv-keys 409B6B1796C275462A17031138
 RUN \curl -sSL https://get.rvm.io | bash -s stable
 
 # Install Ruby 2.5
-RUN apt-get update && apt-get install -y procps && apt-get clean
+RUN apt-get --allow-releaseinfo-change update && apt-get install -y procps && apt-get clean
```
wanlin31, line 33: Wondering why we want to use this option, is there an update error? Would this option install things with an unverified version?

https://manpages.debian.org/unstable/apt/apt-get.8.en.html
https://manpages.debian.org/unstable/apt/apt-secure.8.en.html

containers/pre_built_workers/php7/Dockerfile
```diff
@@ -87,11 +85,10 @@ RUN gpg2 --keyserver keyserver.ubuntu.com --recv-keys 409B6B1796C275462A17031138
 RUN \curl -sSL https://get.rvm.io | bash -s stable
 
 # Install Ruby 2.5
+RUN apt-get --allow-releaseinfo-change update && apt-get install -y procps && apt-get clean
```
wanlin31, line 88: I see the `procps` was installed earlier in line #80, can I ask is there any specific reason to install it again? Failure on previous installation?

containers/runtime/php7/Dockerfile
```diff
@@ -30,12 +30,10 @@ RUN gpg2 --keyserver keyserver.ubuntu.com --recv-keys 409B6B1796C275462A17031138
 RUN \curl -sSL https://get.rvm.io | bash -s stable
 
 # Install Ruby 2.5
-RUN apt-get update && apt-get install -y procps && apt-get clean
+RUN apt-get --allow-releaseinfo-change update && apt-get install -y procps && apt-get clean
 RUN /bin/bash -l -c "rvm install ruby-2.5"
 RUN /bin/bash -l -c "rvm use --default ruby-2.5"
 RUN /bin/bash -l -c "echo 'gem: --no-document' > ~/.gemrc"
-RUN /bin/bash -l -c "echo 'export PATH=/usr/local/rvm/bin:$PATH' >> ~/.bashrc"
-RUN /bin/bash -l -c "echo 'rvm --default use ruby-2.5' >> ~/.bashrc"
```
wanlin31, line 38: On these two line, just want to make sure we still run the right version of ruby. :)

ybbbby: Verified the ruby version is still 2.5 which is what we want.

containers/pre_built_workers/php7/Dockerfile
ybbbby: Right, every time we use `rvm` it requires to `apt-get update`. Otherwise an error will pop up. ref:https://stackoverflow.com/questions/23650992/ruby-rvm-apt-get-update-error

containers/init/build/php7/Dockerfile
ybbbby: I add the `--allow-releaseinfo-change` mainly b/c some of the dependencies have new releases. Several error might pop out if we don't add this, b/c cache are used in previous steps. I won't worry too much on unverified version. Normally, when we use apt-get install, we would install the latest version.

PR #224, opened 2021-10-01
Add support for test case properties

brandonpaiz:


tools/runner/xunit/entities.go
```diff
@@ -133,3 +136,18 @@ type Error struct {
...
+
+// Property encapsulates metadata regarding a test property.
+type Property struct {
+	XMLName xml.Name `xml:"property"`
+	Key     string   `xml:"name,attr"`
+	Value   string   `xml:"value,attr"`
+}
+
+// Sort the properties of a testcase alphabetically by key.
+func (ts *TestCase) sortProperties() {
```
ybbbby, line 148: I would use `tc` as the variable name.

ybbbby (approved):
In general, looks good! One minor comment.

tools/runner/reporter.go
```diff
@@ -217,6 +217,18 @@ func (tcr *TestCaseReporter) Duration() time.Duration {
 	return tcr.endTime.Sub(tcr.startTime)
 }
 
+// AddProperty adds a key-value property to the test case.
+func (tcr *TestCaseReporter) AddProperty(key, value string) {
+	property := &xunit.Property{
```
paulosjca, line 222: You can make this shorter by testing for nil in the beginning and then appending the property in one line instead of creating a separate "property" variable.

PR #226, opened 2021-10-05
Refactor code to allow saving logs with xml reports.

paulosjca:
This change moves the logic to create output directories for xml reports before the tests run, so the runner can save logs to the same locations.


paulosjca:
PTAL @brandonpaiz

brandonpaiz (approved):

PR #227, opened 2021-10-08
Store pod logs

brandonpaiz:


tools/cmd/runner/main.go
```diff
@@ -91,9 +91,10 @@ func main() {
 	done := make(chan *runner.TestSuiteReporter)
 
 	for qName, configs := range configQueueMap {
+		podLogDir := path.Join(path.Dir(o), "pod-logs")
```
paulosjca, line 94: Please delete this line.

tools/cmd/runner/main.go
```diff
@@ -91,9 +91,10 @@ func main() {
...
 		testSuiteReporter := reporter.NewTestSuiteReporter(qName, logPrefixFmt, runner.TestCaseNameFromAnnotations("scenario"))
 		testSuiteReporter.SetStartTime(time.Now())
-		go r.Run(ctx, configs, testSuiteReporter, c[qName], done)
+		go r.Run(ctx, configs, testSuiteReporter, c[qName], podLogDir, done)
```
paulosjca, line 97: Here we should have `outputDirMap[qName]` instead of `podLogDir`.

tools/runner/runner.go
```diff
@@ -59,11 +61,12 @@ func NewRunner(loadTestGetter clientset.LoadTestGetter, afterInterval func(), re
 		afterInterval:         afterInterval,
 		retries:               retries,
 		deleteSuccessfulTests: deleteSuccessfulTests,
+		podLogger:             NewPodLogger(),
 	}
 }
 
 // Run runs a set of LoadTests at a given concurrency level.
-func (r *Runner) Run(ctx context.Context, configs []*grpcv1.LoadTest, suiteReporter *TestSuiteReporter, concurrencyLevel int, done chan<- *TestSuiteReporter) {
+func (r *Runner) Run(ctx context.Context, configs []*grpcv1.LoadTest, suiteReporter *TestSuiteReporter, concurrencyLevel int, podLogDir string, done chan<- *TestSuiteReporter) {
```
paulosjca, line 69: I think the argument here should be `outputDir` instead of `podLogDir`, to be more generic.

paulosjca (changes requested):

tools/runner/runner.go
brandonpaiz: Sounds good, I also changed the argument of the same name for runTest

tools/runner/logger.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	// Attempt to create directory. Will not error if directory already exists
+	err = os.MkdirAll(podLogDir, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("Failed to create pod log output directory %s: %v", podLogDir, err)
+	}
+
+	// Write logs to files
+	errorOccured := false
```
paulosjca, line 41: "Occurred".

I think we should consider fetching the log to memory and then writing it if it is not empty. Logs in our case are small (or empty), and this would separate two classes of errors, one from getting the logs from the pods, and the other from writing the logs to the files.

tools/runner/logger.go
```diff
@@ -0,0 +1,98 @@
...
+// PodLogger provides functionality to save pod logs to files.
+type PodLogger struct {
+}
+
+// NewPodLogger creates a new PodLogger object.
+func NewPodLogger() *PodLogger {
+	return &PodLogger{}
+}
+
+// savePodLogs saves pod logs to files with same name as pod.
```
paulosjca, line 26: I would make a distinction between functions that are part of the interface, such as savePodLogs, and functions that are purely internal. I would rename this function SavePodLogs for this reason. We could move the LogSaver into a subpackage, instead of keeping it in runner, and this would still be valid.

tools/runner/logger.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	if errorOccured {
+		return errors.New("One or more log files could not be written")
+	}
+	return nil
+}
+
+// getTestPods retrieves the pods associated with a LoadTest.
+func (pl *PodLogger) getTestPods(ctx context.Context, loadTest *grpcv1.LoadTest) ([]*corev1.Pod, error) {
+	clientset := getGenericClientset()
```
paulosjca, line 57: getGenericClientSet is really NewGenericClientSet, so I would make the clientset part of the PodLogger/LogSaver structure.

I would also check if the grpcClientSet can be used to get CoreV1().Pods(...). We probably need only NamespaceDefault in any case (our loadtest pods are always in default namespace).

brandonpaiz: The GRPCTestClientSet as it was could not be used to get Pods (as far as I can tell). I modified the interface slightly to provide this functionality. Do you think the solution works? Relevant code in `clientset/client.go`.  

This would make the function `NewGenericClientset`/`getGenericClientset` unnecessary.

tools/runner/logger.go
```diff
@@ -0,0 +1,98 @@
...
+	"path/filepath"
+
+	grpcv1 "github.com/grpc/test-infra/api/v1"
+	"github.com/grpc/test-infra/status"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// PodLogger provides functionality to save pod logs to files.
+type PodLogger struct {
```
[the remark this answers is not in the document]
brandonpaiz: Agreed, I changed the name to LogSaver, and it now stores the clientset

tools/runner/client.go
```diff
@@ -48,6 +49,32 @@ func NewLoadTestGetter() clientset.LoadTestGetter {
 		return nil
 	})
 
+	config := getConfig()
```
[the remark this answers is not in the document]
brandonpaiz: Sounds good, done

clientset/client.go
```diff
@@ -18,6 +18,7 @@ package v1
 
 import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/kubernetes"
```
paulosjca, line 21: I might keep this file unchanged, and then import the kubernetes clientset in the runner client instead (see that file).

tools/cmd/runner/main.go
```diff
@@ -76,7 +76,8 @@ func main() {
 	log.Printf("Queue concurrency levels: %v", c)
 	log.Printf("Output directories: %v", outputDirMap)
 
-	r := runner.NewRunner(runner.NewLoadTestGetter(), runner.AfterIntervalFunction(p), retries, deleteSuccessfulTests)
+	grpcClientset := runner.NewGRPCTestClientSet()
```
paulosjca, line 79: Here we are really want to create a LoadTestGetter and a PodsGetter. Perhaps it would be better to have the grpcClientset be separate from a k8sClientset and each of them to create their respective getter.

tools/runner/client.go
```diff
@@ -48,6 +48,26 @@ func NewLoadTestGetter() clientset.LoadTestGetter {
...
 
+	schemebuilder.AddToScheme(clientgoscheme.Scheme)
+	scheme := clientgoscheme.Scheme
+	types := scheme.AllKnownTypes()
+	_ = types
+
+	return clientset.LoadTestV1().LoadTests(corev1.NamespaceDefault)
+}
+
+// NewGRPCTestClientSet returns a new GRPCTestClientset.
```
paulosjca, line 59: Here I would separate NewGRPCTestClientset (note lowercase s) fom NewK8sClientset, with the second one returning a vanilla client set. I would import k8s.io/kubernetes/clientset as k8sclientset to achieve that.

Thinking about it more, perhaps the interfaces we want are

func NewLoadTestGetter(namespace corev1.Namespace) clientset.LoadTestGetter

We could leave the clientset New methods as they are or build them into the two functions that we really want to use.

and

func NewPodsGetter(namespace corev1.Namespace) corev1.PodsGetter

tools/runner/logsaver.go
```diff
@@ -0,0 +1,110 @@
...
+	grpcv1 "github.com/grpc/test-infra/api/v1"
+	clientset "github.com/grpc/test-infra/clientset"
+	"github.com/grpc/test-infra/status"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LogSaver provides functionality to save pod logs to files.
+type LogSaver struct {
+	clientset clientset.GRPCTestClientset
```
paulosjca, line 21: This might becomre podsGetter corev1.PodsGetter

paulosjca (changes requested):

clientset/client.go
brandonpaiz: Sounds good

tools/runner/logsaver.go
brandonpaiz: Oh yes, I really like the idea of giving the LogSaver only what it needs

tools/cmd/runner/main.go
brandonpaiz: PTAL at main.go now...is this what you mean?

tools/runner/client.go
brandonpaiz: I separated the clientset functions (and changed the s).

I went ahead and renamed it to NewGRPCTestClientset. It now gets the grpc clientset itself by calling NewGRPCTestClientset.

It looks to me like a namespace argument in NewPodsGetter would be unused, so I did not add this. Am I misunderstanding something?
For NewLoadTestGetter, the namespace is used but because (1) the namespace never changes and (2) this function is called from main and would add unnecessary imports/complexity, I did not add this argument. But again please let me know if I am confusing something.

PR #229, opened 2021-10-29
Configuration to setup prometheus.

wanlin31:
This commit takes some of the configuration from Istio/Tools to
set up prometheus and servicemonitor for scraping cpu and memory
usage.

Steps to generate the configuration:
1. Following script [tools/perf/istio-install/setup_istio.sh line 95](https://github.com/istio/tools/blob/22fa6c5a036f9de09d31f8744bdcd728995ffa7d/perf/istio-install/setup_istio.sh#L95)
2. Leave the part related to Istio out, only take the portion related to prometheus operator and creating service accounts, clusterroles and clusterrolebinding.
3. Change the namespaces.

### DIFF
--- a/config/prometheus/setup/README.md
+++ b/config/prometheus/setup/README.md
@@ -1,0 +1,10 @@
+The prometheus and servicemonitor configurations are taken from 
+[istio/tools](https://github.com/istio/tools).
+
+Steps to generate the configuration:
+
+* The base configurations and script used came from 
+[tools/perf/istio-install/](https://github.com/istio/tools/tree/master/perf/istio-install#istio-setup)
+* Leave the part related to istio out, only take the portion related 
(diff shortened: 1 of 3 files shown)


wanlin31:
I have wrote a README.md to say that these configurations are mostly generated using istio/tools repo and scraped away parts related to istio, anything else should I do to make sure the logistics are okay?

In the configuration, the namespace is fixed as `prometheus`, that requires the namespace: `prometheus` being created before deploying the prometheus services. We can also use helm to fill in the namespace, but that is an bit over in my opinion since: 
* `namespace` is the only field need to be filled in.
*  Create `prometheus` namespace can be part of the cluster, no extra burden

paulosjca (approved):

PR #230, opened 2021-11-01
Add pod-viewer-role to kustomization.yaml

wanlin31:
This commit create pod-viewer-role when the role did not exit.

### DIFF
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -6,9 +6,9 @@ resources:
 - leader_election_role.yaml
 - leader_election_role_binding.yaml
 - loadtest_editor_role.yaml
+- loadtest_component_bindings.yaml
 - loadtest_viewer_role.yaml
+- manager_binding.yaml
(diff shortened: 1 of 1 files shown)


config/rbac/kustomization.yaml
```diff
@@ -7,6 +7,7 @@ resources:
 - leader_election_role_binding.yaml
 - loadtest_editor_role.yaml
 - loadtest_viewer_role.yaml
+- pod_viewer_role.yaml
 - role.yaml
 - manager_binding.yaml
 - loadtest_component_bindings.yaml
```
ybbbby, line 13: Since we are here, we might want to make the whole list alphabetically :)

ybbbby (approved):
LGTM

config/rbac/kustomization.yaml
wanlin31: I will do that :)

paulosjca (approved):

PR #231, opened 2021-11-02
Use nproc to get processor count

brandonpaiz:


containers/pre_built_workers/java/run_worker.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# Copyright 2020 gRPC authors
```
paulosjca, line 2: Since this is a new file, 2021 would be appropriate.

brandonpaiz:
Corresponding PR https://github.com/grpc/grpc/pull/27919

paulosjca (commented):
Don't you need similar changes to the examples, and perhaps also to the load test configuration templates in grpc/grpc? If you don't change the templates, this script will not be called. But then this could be done in follow-up PRs.

You should update the basic examples in a separate PR just before the next release of the controller, and update the basic templates in grpc/grpc once the release is complete.

containers/pre_built_workers/java/run_worker.sh
brandonpaiz: Updated 👍

paulosjca (commented):
There should be a Dockerfile for non-prebuilt workers as well, since we are already updating the examples.

paulosjca (approved):

jtattermusch:
Thanks for the fix. Please fill in the PR description to describe the context in a sentence or two.

PR #232, opened 2021-11-04
Support timeout when building images

jtattermusch:
Currently, some kokoro runs of OSS benchmarks get stuck forever and they are only killed by a top level kokoro job timeout (after X hours). I've looked into that and discovered that the PHP7 image build sometimes get stuck forever.
The exact reason why PHP7 prebuilt image built gets stuck can be investigated later (it's quite a weird reason AFAICT), but the framework itself need to be robust enough and kill stuck builds after a while (and print some useful output that allows us to see what it is that got stuck).

Also note that the stuck CI jobs taking 4hours until they get killed by kokoro are the reason why we see intermittent gaps in our dashboard's data.

Examples of stuck builds (you'll see that the logs aren't very helpful)
https://source.cloud.google.com/results/invocations/4689a61b-8a04-401a-9bf6-f7f1dc0b62d0
https://source.cloud.google.com/results/invocations/751e942d-0336-4d21-b367-de59ea2f5199
https://source.cloud.google.com/results/invocations/49d8c480-4c01-4235-9058-7f07d28e57cf

Note that my golang skills are basically nonexistent, so feel free to suggest a better way of implementing timeouts. But what I currently have is good as a hotfix I guess.
(I think ideally we'd have a function that executes a command with predefined timeout, tracks the status of the command e.g. by printing current status periodically and dumps the output in a readable fashion when it succeeds/fails).

### DIFF
--- a/tools/cmd/prepare_prebuilt_workers/main.go
+++ b/tools/cmd/prepare_prebuilt_workers/main.go
@@ -137,14 +137,18 @@ func main() {
 
 			// build image
 			log.Println(fmt.Sprintf("building %s image", lang))
-			buildDockerImage := exec.Command("docker", "build", dockerfileLocation, "-t", image, "--build-arg", fmt.Sprintf("GITREF=%s", gitRef), "--build-arg", fmt.Sprintf("BREAK_CACHE=%s", uniqueCacheBreaker))
+			// TODO(jtattermusch): allow setting REPOSITORY arg to allow building from user forks e.g. https://github.com/USER_NAME/grpc.git
+			buildCommandTimeoutSeconds := 30 * 60 // 30 mins should be enough for all languages
+			buildDockerImage := exec.Command("timeout", fmt.Sprintf("%ds", buildCommandTimeoutSeconds), "docker", "build", dockerfileLocation, "-t", image, "--build-arg", fmt.Sprintf("GITREF=%s", gitRef), "--build-arg", fmt.Sprintf("BREAK_CACHE=%s", uniqueCacheBreaker))
 			buildOutput, err := buildDockerImage.CombinedOutput()
 			if err != nil {
-				log.Println(err)
-				log.Fatalf("Failed building %s image: %s", lang, string(buildOutput))
+				log.Printf("Failed building %s image. Dump of command's output will follow:\n", lang)
+				log.Println(string(buildOutput))
+				log.Fatalf("Failed building %s image: %s", lang, err.Error())
 			}
+			log.Printf("Succeeded building %s image. Dump of command's output will follow:\n", lang)
 			log.Println(string(buildOutput))
-			log.Printf("Succeeded building %s worker: %s\n", lang, image)
+			log.Printf("Succeeded building %s image: %s\n", lang, image)
 
 			if !test.buildOnly {
 				// push image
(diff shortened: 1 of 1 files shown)


wanlin31 (approved):

tools/cmd/prepare_prebuilt_workers/main.go
```diff
@@ -137,26 +137,32 @@ func main() {
 
 			// build image
 			log.Println(fmt.Sprintf("building %s image", lang))
-			buildDockerImage := exec.Command("docker", "build", dockerfileLocation, "-t", image, "--build-arg", fmt.Sprintf("GITREF=%s", gitRef), "--build-arg", fmt.Sprintf("BREAK_CACHE=%s", uniqueCacheBreaker))
+			// TODO(jtattermusch): allow setting REPOSITORY arg to allow building from user forks e.g. https://github.com/USER_NAME/grpc.git
+			buildCommandTimeoutSeconds := 30 * 60 // 30 mins should be enough for all languages
```
wanlin31, line 141: It should be enough to finish any language on kokoro machines. I am okay with a hard coded timeout, just wondering if it would be useful to make it a flag with default value.

tools/cmd/prepare_prebuilt_workers/main.go
```diff
@@ -137,26 +137,32 @@ func main() {
...
 			buildOutput, err := buildDockerImage.CombinedOutput()
 			if err != nil {
-				log.Println(err)
-				log.Fatalf("Failed building %s image: %s", lang, string(buildOutput))
+				log.Printf("Failed building %s image. Dump of command's output will follow:\n", lang)
+				log.Println(string(buildOutput))
+				log.Fatalf("Failed building %s image: %s", lang, err.Error())
 			}
+			log.Printf("Succeeded building %s image. Dump of command's output will follow:\n", lang)
 			log.Println(string(buildOutput))
```
wanlin31, line 150: In here we just dump all the buildOutput if it was built successfully , it is pretty lengthy, should we only print these build out put when an error occurred?

wanlin31:
I agree we should have a timeout to bound the building time.

I think use `timeout` there should work fine. We may also use `exec.CommandContext` instead of `exec.Command` to set up context WithTimeout. I think both of them should work.

https://pkg.go.dev/os/exec#CommandContext
https://pkg.go.dev/context#WithTimeout

paulosjca (approved):

tools/cmd/prepare_prebuilt_workers/main.go
jtattermusch: I will leave as is in this PR to get the fix in ASAP, but feel free to address as a followup PR.

tools/cmd/prepare_prebuilt_workers/main.go
jtattermusch: Yes, that would work, we actually do something similar in our test scripts in grpc/grpc.
But it has downsides as well - sometimes is just useful to be able to inspect the log for successful builds.

I think the best solution is to make sure that the logs are stored in .log files and that there's a corresponding sponge_log.xml  file. If done correctly, one could see the structured results in the resultstore UI, and be able to see the full "target log" in the resultstore UI (but the main log wouldn't be polluted by the output).
But this solution basically requires building a "task runner" method that does all this for you. In grpc/grpc we have something like this written in python, so I don't know if it makes sense to recreate it in go.

jtattermusch:
> I agree we should have a timeout to bound the building time.
> 
> I think use `timeout` there should work fine. We may also use `exec.CommandContext` instead of `exec.Command` to set up context WithTimeout. I think both of them should work.
> 
> https://pkg.go.dev/os/exec#CommandContext https://pkg.go.dev/context#WithTimeout

I considered that but then realized my go skills aren't very good and something like this would also require tests. 
Btw, basically this is go version of https://github.com/grpc/grpc/blob/master/tools/run_tests/python_utils/jobset.py (which handles running multiple jobs in parallel, printing the results, generating the xml reports etc.)

PR #233, opened 2021-11-04
Cache RVM build keys to reduce flakiness.

paulosjca:
This change caches the public key required to build PHP7 images in a file, to reduce flakiness. Three copies of the file are stored, since there are three images built in different contexts.

This change includes #234.


jtattermusch:
Thanks for the attempt, but I think we should understand why the command gets stuck in the first place (because it really shouldn't), and then address that root cause. Applying a dirty workaround without really understanding the problem doesn't seem like a good solution in this case.

paulosjca:
> Thanks for the attempt, but I think we should understand why the command gets stuck in the first place (because it really shouldn't), and then address that root cause. Applying a dirty workaround without really understanding the problem doesn't seem like a good solution in this case.

I propose to start with #234, and move to this one if flakiness reoccurs. This PR removes the command in question, so  it seems to me like an appropriate fix. #234 may be sufficient, since it gives some insight that the issue is related to the base docker image, and brings our build into line with the interop build (which, as far as I can tell, is not flaky).

jtattermusch:
> > Thanks for the attempt, but I think we should understand why the command gets stuck in the first place (because it really shouldn't), and then address that root cause. Applying a dirty workaround without really understanding the problem doesn't seem like a good solution in this case.
> 
> I propose to start with #234, and move to this one if flakiness reoccurs. This PR removes the command in question, so it seems to me like an appropriate fix. #234 may be sufficient, since it gives some insight that the issue is related to the base docker image, and brings our build into line with the interop build (which, as far as I can tell, is not flaky).

Let's go with #234.

The problem I have with this fix is that it's basically just expanding on the scope of this TODO: https://github.com/grpc/test-infra/blob/777dba9391223dc913cdb43088b95e9e6d25c205/containers/pre_built_workers/php7/Dockerfile#L17

Ideally, we wouldn't be running any of the commands on these lines: https://github.com/grpc/test-infra/blob/777dba9391223dc913cdb43088b95e9e6d25c205/containers/pre_built_workers/php7/Dockerfile#L17-L42 on each build. Ideally we'd have a base docker image that already has all these prerequisites installed, and we would only be running the actual build commands.
That would also reduce the duplication between the pre_build, build and runtime docker images.
Similar duplication exists not just for PHP, but also for other languages (e.g. C#).

PR #234, opened 2021-11-08
Update PHP7 version for OSS benchmarks images.

paulosjca:
* Updates PHP7 image versions from php7.2.22-buster to php7.2.34-buster (latest is 7.4.25).
* Updates gpg2 invocation to exclude unused key and remove hardcoded key server specification.
* Proposed as an alternative to #233.

### DIFF
--- a/containers/init/build/php7/Dockerfile
+++ b/containers/init/build/php7/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM php:7-buster
+FROM php:7.2.34-buster
 RUN mkdir -p /src/workspace
 WORKDIR /src/workspace
 
--- a/containers/init/build/php7/Dockerfile
+++ b/containers/init/build/php7/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM php:7.2.22-buster
+FROM php:7-buster
 RUN mkdir -p /src/workspace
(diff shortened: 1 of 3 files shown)


containers/init/build/php7/Dockerfile
```diff
@@ -21,12 +21,12 @@ RUN apt-get update && apt-get install -y \
...
   lcov \
-  make && \
+  make \
+  gnugp2 && \
   apt-get clean
 
 # Install rvm
-RUN apt-get update && apt-get install -y gnupg2 && apt-get clean
-RUN gpg2 --keyserver keyserver.ubuntu.com --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
+RUN gpg2 --recv-keys 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
```
ybbbby, line 29: This modification would make gpg2 ask embedded key server for public key?

paulosjca: Yes this is the same as what happens for interop.

https://github.com/grpc/grpc/blob/73dcfd375d14c7cef537a5255767a8950424988c/tools/dockerfile/interoptest/grpc_interop_ruby/Dockerfile#L73

paulosjca:
I propose to merge this PR and, if flakiness should reoccur, move to #233.

jtattermusch (approved):
LGTM. Thanks for the fix. I'm wondering if this is actually going to help though. I guess we will see soon enough.

PR #235, opened 2021-11-08
Implementation of xds server

wanlin31:
The PR implement a the xDS server using the config library implemented in PR 245. The xDS server implemented here is expected to run in a container, the main process would start with building resource snapshot from the configuration json file and start a resource update server to listen for the podIP and port from the ready container, the process will be block until the podIP and port information is received. Upon receiving,  the resource snapshot generated from the configuration file will be updated and the xDS server will be started server the resource.


wanlin31:
The PR is a working solution but still half way through, want to have an early discussion on following items: 
1. To limit the resource that xDS server consume, in the main process, I set it to only listen to the update once and shut down the update server afterwards. One concern maybe if the the node that the server pod running on dies, the pod will be recreated and lost its IP, in that we can make the test fail, is that acceptable? 
2. Currently the resources served are hardcoded in the library, for the client side tests, do we want to make the resources configurable? i.e. building resource from a configuration file
3. I have not yet add any tests, they seems trivial to me to test, but wanted to learn more opinions.

wanlin31:
The code are ready for review, still working on the README.md, would appreciated any thoughts.

containers/runtime/xds/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN mkdir -p /src/workspace
+WORKDIR /src/workspace
+
+# when BREAK_CACHE arg is set to a random value (e.g. by "--build-arg BREAK_CACHE=$(uuidgen)"),
+# it makes sure the docker cache breaks at this command, and all the following
+# commands in this Dockerfile will be forced to re-run on each build.
+# This is important to ensure we always clone the repository even if "GITREF" stays unchanged
+# (important e.g. when GITREF=master, when the clone command could get cached and
+# we'd end up with a stale repository).
+ARG BREAK_CACHE
```
gnossen, line 12: Would the [`--no-cache`](https://docs.docker.com/engine/reference/commandline/build/#:~:text=instructions%20during%20build-,%2D%2Dno%2Dcache,-Do%20not%20use) arg to `docker build` work here?

containers/runtime/xds/cmd/main.go
```diff
@@ -0,0 +1,97 @@
...
+	flag.StringVar(&nodeID, "nodeID", "test_id", "Node ID")
+
+	// Default configuration path, the path is relative path using ./containers/runtime/xds
+	flag.StringVar(&defaultConfigPath, "d", "containers/runtime/xds/config/default_config.json", "The path of default configuration file, the path is relative path the root of test-infra repo")
+
+	// User supplied configuration path, the path is relative path using ./containers/runtime/xds
+	flag.StringVar(&userSuppliedConfigPath, "u", "", "The path of user supplied configuration file, the path is relative path the root of test-infra repo")
+
+	// Tne cluster name for Envoy obtain configuration from, should match the cluster name in the bootstrap file.
+	flag.StringVar(&resource.XDSServerClusterName, "c", "xds_cluster", "Tne cluster name for Envoy to obtain configuration, should match the cluster name in the bootstrap file")
```
gnossen, line 48: I don't think I understand what's going on here. Why is there a static cluster in the Envoy bootstrap? Also, this flag doesn't seem to be used anywhere.

containers/runtime/xds/cmd/main.go
```diff
@@ -0,0 +1,97 @@
...
+	log.Println("xDS server resource snapshot is generated successfully")
+
+	// Create a cache
+	l := logrus.New()
+	cache := cache.NewSnapshotCache(false, cache.IDHash{}, l)
+
+	// Start the endpoint update server
+	endpointChannel := make(chan []*config.TestEndpoint)
+
+	go func() {
```
gnossen, line 74: Nit: Why wrap in an anonymous function? Couldn't this just be
```
go xds.RunUpdateServer(endpointChannel, endpointUpdaterPort)
```

containers/runtime/xds/cmd/main.go
gnossen: Ah, after looking at https://github.com/grpc/test-infra/pull/242, I see why you need this. But I'm not sure why the xDS server needs to know this cluster name? Is it just because you need to avoid a collision in cluster names?

wanlin31: If we use ads{} for LDS/RDS and CDS/EDS we would not need the cluster name, but we used the cluster name in RDS configSource: ["clusterName": "xds_cluster"](https://github.com/grpc/test-infra/blob/10f8de6bcbaa9377267dff804d9c470ae0da4281/containers/runtime/xds/config/default_config.json#L115), I will worker further on using ads for RDS for Envoy, if not possible I will double check to see if this field is required.  

My understanding is APIserver is only used by gRPC proxyless clients and may not be used by Envoy, so there are two listeners pointing to the same cluster/endpoint. Preferably the two listeners should be more identical, the api_listener (proxyless clients) is using ads, I will try to make the Envoy listener able to use ads{} as well.

containers/runtime/xds/Dockerfile
wanlin31: I think `--no-cache` should work great here, since we really don't have much things to keep before installing copy step, always starting from scratch makes scene. Thank you for letting me know!

containers/runtime/xds/cmd/main.go
wanlin31: Fixed.

containers/runtime/xds/cmd/main.go
wanlin31: I have tried to use ads for RSD config, did not work properly, I think it will take a bit to figure out. Before How important would it be to use the same type of config source (ads vs rds) for Envoy and gRPC xds client?

wanlin31: @gnossen I have tested that we are able to use ads for fetching all resources, if we are sure we always want to use ads we may remove this field. Do you think we may want to keep this field in case in the future someone want to fetch rds or eds separately? Same thing goes with the servers, if we are sure we always use ads we can register the ads server only.

gnossen: Let's stick with ADS. I find it highly unlikely that someone will want to configure that option in these benchmarks.

wanlin31: I will update the code to only use ADS.

gnossen:
@wanlin31 I see that https://github.com/grpc/test-infra/pull/255 includes the entirety of the contents of this PR. I have already done an initial round of review on that PR. Please resolve all of the appropriate comments on that PR and we'll continue the rest of the review for the content of this PR on this PR.

wanlin31:
@gnossen Right, #255 has to depend on #235, I am sorry for the confusion, this is the first time I encountered this situation, how can I better annotate this PR so that it won't cause this type of confusion any more? 

I will address comments that belongs to the PR here and the ones belong to #255 in 255. Thank you.  😊

gnossen (approved):

wanlin31:
The whole pr was included in the #268

PR #236, opened 2021-11-08
Add loadtest name as property

brandonpaiz:


tools/runner/runner.go
```diff
@@ -145,6 +145,7 @@ func (r *Runner) runTest(ctx context.Context, config *grpcv1.LoadTest, reporter
 			if err != nil {
 				reporter.Error("Could not save pod logs: %s", err)
 			}
+			reporter.AddProperty("Load test name", loadTest.Name)
```
wanlin31, line 148: Wondering if we want to use LoadTest name. :)

brandonpaiz: Yes, that sounds good :)

tools/runner/runner.go
```diff
@@ -145,6 +145,7 @@ func (r *Runner) runTest(ctx context.Context, config *grpcv1.LoadTest, reporter
 			if err != nil {
 				reporter.Error("Could not save pod logs: %s", err)
 			}
+			reporter.AddProperty("LoadTest name", loadTest.Name)
```
paulosjca, line 148: Properties are keys, so it is better to treat them like identifiers and not like free text.

In particular, we want to make sure that properties are grouped together and identifiable as part of a group. Keeping them short is something that helps with that too.

I suggest calling this one simply `name`.

The properties related to pods will then be something like `pod.client-0.name`, `pod.client-0.log` and so on (`pod` or `pods` will work). The reason is that when you sort them alphabetically the properties of pods are grouped together, and the properties of each pod are grouped together (if we had something like `log.pod.client-0` this wouldn't work).

brandonpaiz: Alright, this sounds good. Change made.

paulosjca (approved):

PR #237, opened 2021-11-09
Update dotnet SDK and runtime in C# build

jtattermusch:
This will be needed for https://github.com/grpc/grpc/pull/27966,
but since the build definitions for C# live in two repositories (grpc/grpc and grpc/test-infra), it's difficult to change the C# build atomically and cleanly.

The current plan is that once both PRs are approved, I'll merge them in short succession.

### DIFF
--- a/containers/pre_built_workers/csharp/Dockerfile
+++ b/containers/pre_built_workers/csharp/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM mcr.microsoft.com/dotnet/sdk:2.1
+FROM mcr.microsoft.com/dotnet/sdk:6.0-bullseye-slim
 
 # TODO: when running on kokoro, this build step will not be cached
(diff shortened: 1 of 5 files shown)


jtattermusch:
I tested the image build together with grpc/grpc#27966 manually and it seems to be working fine.

containers/pre_built_workers/csharp/Dockerfile
```diff
@@ -49,7 +49,10 @@ RUN mkdir /build_scripts
 ADD build_qps_worker.sh /build_scripts
 RUN /build_scripts/build_qps_worker.sh
 
-FROM mcr.microsoft.com/dotnet/runtime:2.1
+# Note that the libgrpc_csharp_ext.so library built by the build image
+# needs to binary compatible with the runtime image (e.g. they need to be based off of the same linux distro)
```
paulosjca, line 53: "needs to be binary-compatible"; other than that, LGTM.

paulosjca (approved):

PR #238, opened 2021-11-09
Add pod to log filepath map

brandonpaiz:
This change enables the runner to use the map retuned by SavePodLogs to add pod names as properties in the test output.

### DIFF
--- a/tools/runner/logsaver.go
+++ b/tools/runner/logsaver.go
@@ -28,8 +28,9 @@ func NewLogSaver(podsGetter corev1types.PodsGetter) *LogSaver {
 	}
 }
 
-// SavePodLogs saves pod logs to files with same name as pod. It returns a map
-// where pods are keys and values are the filepath of the saved log.
+// SavePodLogs saves pod logs to files with same name as pod.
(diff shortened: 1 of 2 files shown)


tools/runner/logsaver.go
```diff
@@ -28,33 +28,37 @@ func NewLogSaver(podsGetter corev1types.PodsGetter) *LogSaver {
 	}
 }
 
-// SavePodLogs saves pod logs to files with same name as pod.
-func (ls *LogSaver) SavePodLogs(ctx context.Context, loadTest *grpcv1.LoadTest, podLogDir string) error {
+// SavePodLogs saves pod logs to files with same name as pod. It returns a map
```
paulosjca, line 31: The first line in the godoc should stand by itself, because it is extracted for documentation.

Breaking after the period and writing something like "This function returns a map..." should be enough.

You may want to keep these lines within 80 columns.

brandonpaiz: Done. I'm going to spend a few minutes looking at godoc so I understand how docs are generated 👍

paulosjca (approved):

PR #239, opened 2021-11-11
Add protofile for endpoint updater

wanlin31:
This commit add the proto file to generate the gRPC service for communicate the test server's IP address.

The test server's pod IP will be sent from driver's ready container to xDS server container(on the client pod).


wanlin31:
Is the service name okay, seems long for me but I can't find a better name.  

Also where to put these files? 
I believe these files should belongs to test-infra, the service would be used in a container that deployed with the test client, I don't expect too much changes on the xDS server implementation so the image can be stable enough to be built with test-infra's release. If the file is in gRPC/gRPC the implementation of xds server would also be in gRPC, that may cause us to prebuilt the xds server each run.

proto/endpointupdater/endpoint.proto
```diff
@@ -0,0 +1,32 @@
...
+  // Sends an update
+  rpc UpdateEndpoint (EndpointUpdaterRequest) returns (EndpointUpdaterReply) {}
+}
+
+message EndpointUpdaterRequest {
+  string ip_address = 1;
+}
+
+message EndpointUpdaterReply {
+  string error_reason = 1;
```
gnossen, line 31: It's more conventional to use [status](https://pkg.go.dev/google.golang.org/grpc/status) to communicate errors.

wanlin31: I am not very familiar with how to use the status, can you give me a reference to an example? 

I am assuming that the status is something we can check on the client side when we send out the request, if that is the case, would we be return nothing from the service? Thank you. :)

gnossen: [This post](https://jbrandhorst.com/post/grpc-errors/) looks like a decent write-up on how to use status in Go.

> I am assuming that the status is something we can check on the client side when we send out the request, if that is the case, would we be return nothing from the service

Exactly. Just an empty message.

wanlin31: I have read the post, good to learn how to handle gRPC error properly. Updated. Thank you!

gnossen (approved):

paulosjca (approved):
Just change the name in the PR to "endpoint updater".

PR #240, opened 2021-11-11
Add pod names to properties

brandonpaiz:
Sample output:
```
<testsuites name="" tests="1" errors="0" time="181.396307675">
  <testsuite id="0" name="" tests="1" errors="0" time="181.396307675">
    <testcase name="java-generic-async-streaming-ping-pong-secure" time="181.396193717">
      <properties>
        <property name="name" value="brandonpaiz-pod-test-010"></property>
        <property name="pod.client-0.name" value="brandonpaiz-pod-test-010-client-0"></property>
        <property name="pod.driver-0.name" value="brandonpaiz-pod-test-010-driver-0"></property>
        <property name="pod.server-0.name" value="brandonpaiz-pod-test-010-server-0"></property>
      </properties>
    </testcase>
  </testsuite>
</testsuites>
```
This is in preparation to add `pod.client.url` / `pod.driver.url` / `pod.server.url` properties

### DIFF
--- a/tools/runner/logsaver.go
+++ b/tools/runner/logsaver.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	grpcv1 "github.com/grpc/test-infra/api/v1"
 	"github.com/grpc/test-infra/status"
@@ -31,19 +32,19 @@ func NewLogSaver(podsGetter corev1types.PodsGetter) *LogSaver {
 // SavePodLogs saves pod logs to files with same name as pod.
 // This function returns a map where pods are keys and values are the filepath
 // of the saved log.
-func (ls *LogSaver) SavePodLogs(ctx context.Context, loadTest *grpcv1.LoadTest, podLogDir string) (map[*corev1.Pod]string, error) {
-	podToLogPath := make(map[*corev1.Pod]string)
+func (ls *LogSaver) SavePodLogs(ctx context.Context, loadTest *grpcv1.LoadTest, podLogDir string) (*SavedLogs, error) {
+	savedLogs := NewSavedLogs()
 
 	// Get pods for this test
 	pods, err := ls.getTestPods(ctx, loadTest)
 	if err != nil {
-		return podToLogPath, err
+		return savedLogs, err
 	}
 
 	// Attempt to create directory. Will not error if directory already exists
 	err = os.MkdirAll(podLogDir, os.ModePerm)
 	if err != nil {
-		return podToLogPath, fmt.Errorf("Failed to create pod log output directory %s: %v", podLogDir, err)
+		return savedLogs, fmt.Errorf("Failed to create pod log output directory %s: %v", podLogDir, err)
 	}
 
 	// Write logs to files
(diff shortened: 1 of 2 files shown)


brandonpaiz:
New sample output:
```
<testsuites name="" tests="1" errors="0" time="181.396307675">
  <testsuite id="0" name="" tests="1" errors="0" time="181.396307675">
    <testcase name="java-generic-async-streaming-ping-pong-secure" time="181.396193717">
      <properties>
        <property name="name" value="brandonpaiz-pod-test-010"></property>
        <property name="pod.client-0.name" value="brandonpaiz-pod-test-010-client-0"></property>
        <property name="pod.driver-0.name" value="brandonpaiz-pod-test-010-driver-0"></property>
        <property name="pod.server-0.name" value="brandonpaiz-pod-test-010-server-0"></property>
      </properties>
    </testcase>
  </testsuite>
</testsuites>
```

tools/runner/runner.go
```diff
@@ -141,11 +141,15 @@ func (r *Runner) runTest(ctx context.Context, config *grpcv1.LoadTest, reporter
 		status = statusString(config)
 		switch {
 		case loadTest.Status.State.IsTerminated():
-			_, err := r.logSaver.SavePodLogs(ctx, loadTest, outputDir)
+			podToPathMap, err := r.logSaver.SavePodLogs(ctx, loadTest, outputDir)
 			if err != nil {
 				reporter.Error("Could not save pod logs: %s", err)
 			}
 			reporter.AddProperty("name", loadTest.Name)
+			for pod := range podToPathMap {
```
paulosjca, line 149: Intricate logic like this should not be here. It should be in its own function. You might even define the "podToPathMap" as its own type and make it a method that returns another map or other data structure containing the key-value pairs for properties. Here you would iterate over that and add them to properties.

tools/runner/runner.go
```diff
@@ -141,11 +141,15 @@ func (r *Runner) runTest(ctx context.Context, config *grpcv1.LoadTest, reporter
...
+				propertyKey := "pod." + strings.TrimPrefix(pod.Name, loadTest.Name+"-") + ".name"
```
paulosjca, line 150: As a secondary note, you probably want to have a function that generates a "pod to property name element" for the pod (just the "client-0" and so on) and functions that generate property name and property log keys from this element. I wouldn't be doing + + but either fmt.Sprintf or strings.Join in those functions, and I wouldn't put them in runner.go either.

tools/runner/logsaver.go
```diff
@@ -113,3 +114,32 @@ func (ls *LogSaver) writeBufferToFile(buffer *bytes.Buffer, filePath string) err
...
+		podToPathMap: make(map[*corev1.Pod]string),
+	}
+}
+
+// GenerateProperties creates pod-log related properties.
+func (sl *SavedLogs) GenerateProperties(loadTest *grpcv1.LoadTest) map[string]string {
+	properties := make(map[string]string)
+	for pod := range sl.podToPathMap {
+		name := sl.podToPropertyName(pod.Name, loadTest.Name, "name")
+		properties[name] = pod.Name
```
brandonpaiz, line 135: Propreties for pod log urls could then be added with something like
```
url := sl.podToPropertyName(pod.Name, loadTest.Name, "url")
properties[name] = calculatedUrl
```

paulosjca (approved):

PR #243, opened 2021-12-01
Update bazel image for driver build.

paulosjca:
Latest image is 3.5.0 and builds without errors.

### DIFF
--- a/containers/runtime/driver/Dockerfile
+++ b/containers/runtime/driver/Dockerfile
@@ -34,8 +34,7 @@ RUN git clone https://github.com/$REPOSITORY.git .
 RUN git submodule update --init
 RUN git checkout $GITREF
 
-# TODO(jtattermusch): why is such and old version of bazel used?
-FROM l.gcr.io/google/bazel:0.17.1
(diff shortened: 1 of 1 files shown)


containers/runtime/driver/Dockerfile
```diff
@@ -34,8 +34,7 @@ RUN git clone https://github.com/$REPOSITORY.git .
 RUN git submodule update --init
 RUN git checkout $GITREF
 
-# TODO(jtattermusch): why is such and old version of bazel used?
-FROM l.gcr.io/google/bazel:0.17.1
+FROM l.gcr.io/google/bazel:3.5.0
```
jtattermusch, line 37: FYI the entire grpc/grpc repo is currently using bazel 4.2.1
https://github.com/grpc/grpc/blob/50a207ee9cc2cf4ffbf3e34f7a46af854b3535f5/tools/bazel#L45

If this version work, it's fine with me (definitely better than the ancient 0.17.0), but it might be worth brining the version in sync with grpc/grpc at some point.

jtattermusch (approved):

PR #244, opened 2021-12-01
Add dashboard code and configuration

brandonpaiz:
Dashboard code is in `test-infra/dashboard`, and its configuration files are in `test-infra/dashboard/config`. Containers are prepared by running `make configure-grafana` or `make configure-replicator`.


cmd/postgres_replicator/README.md
```diff
@@ -0,0 +1,66 @@
...
+## Running
+
+From the test-infra project root, run `make postgres_replicator`, then
+`bin/postgres_replicator -c <config_file>`.
+
+When the postgres replicator receives a `GET` request for `/run`, it will
+transfer new data since the last time it was run. When a transfer is in
+progress, it will ignore additional requests to `/run` (but still return `200`).
+
+
```
jtattermusch, line 62: I think there should be a section that describes how the replication works
(e.g. the automated schema creation, how does the type conversion work etc.).
Also you should mention the client libraries you are using for talking to bigquery and cloudsql.

cmd/postgres_replicator/README.md
```diff
@@ -0,0 +1,66 @@
...
+# Source database settings
+bigQuery:
+  projectID: ${BQ_PROJECT_ID}
+
+# Destination database settings
+postgres:
+  dbHost: 172.17.0.1
+  dbPort: 5432
+  dbUser: ${PG_USER}
+  dbPass: ${PG_PASS}
```
jtattermusch, line 20: can we use https://github.com/GoogleCloudPlatform/cloudsql-proxy?

It looks like you can run cloudsql-proxy locally, have it authenticate to cloudsql instance using application default credentials (https://cloud.google.com/sql/docs/mysql/connect-admin-proxy#authentication-options) and then simply connect to the local cloudsql-proxy endpoint?

postgres_replicator/transfer.go
```diff
@@ -0,0 +1,228 @@
...
+		logger.Errorf("Could not transfer one or more rows to Postgres: %v. ", err)
+		done <- true
+		return
+	}
+
+	done <- true
+}
+
+func (t *Transfer) convertSchema(bqSchema *BigQuerySchema) (*PostgresSchema, error) {
+	pgSchema := &PostgresSchema{make(map[string]string)}
```
jtattermusch, line 120: there should be a comment that explains the logic.
so you take the orignal schema as is and update the data types that are named differently in biqquery?
You should probably also link to the list of supported data types in bigquery and postgres and mention what are all the types that are supported (I really doubt that this would support all the data types, so you should add some explanation in places where you made assumptions).

postgres_replicator/transfer.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (t *Transfer) getBigQueryRows(bigQueryDataset, tableName, dateField string, bqSchema *BigQuerySchema) (*bigquery.RowIterator, error) {
+	// Get most recent entry from Postgres table
+	timestamp, err := t.pg.GetMostRecentEntry(tableName, dateField)
+	if err != nil {
+		return nil, fmt.Errorf("Could not get most recent Postgres timestamp: %s", err)
+	}
+
+	// Get data after this time, or all data if last timestamp doesn't exist
+	rows, err := t.bq.GetDataAfterDatetime(bigQueryDataset, tableName, dateField, timestamp, bqSchema)
```
jtattermusch, line 165: if you're really planning to load all the newly added data in memory (which can be a lot btw) you should at least add a comment/todo why this is ok (you're assuming that this is only used for benchmark results and that there is only going to be a small number of them since you sync frequently).
In general this is not ok (you should at least do updates in batches).

postgres_replicator/transfer.go
```diff
@@ -0,0 +1,228 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"cloud.google.com/go/bigquery"
+	"github.com/leporo/sqlf"
```
jtattermusch, line 11: why was this library chosen? github.com/leporo/sqlf
Without more context it seems arbitrary.

jtattermusch (commented):
I added some initial comments.
I didn't review the golang code in detail (I expect @paulosjca or @wanlin31 to review)

Makefile
```diff
@@ -41,7 +41,7 @@ GOARGS = -trimpath
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-all: controller all-tools
+all: controller all-tools postgres_replicator
```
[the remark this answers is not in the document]
brandonpaiz: Done; I've moved the code into `test-infra/dashboard` and given it it's own makefile.

cmd/postgres_replicator/README.md
```diff
@@ -0,0 +1,66 @@
+# Postgres replicator
+
+This tool replicates streaming data from BigQuery into PostgreSQL.
```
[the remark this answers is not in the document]
brandonpaiz: Maybe the word doesn't fit here, but I meant that the Postgres replicator exclusively operates
on data with stream-like qualities:
1. The data is sequential. For example, rows contain a timestamp column storing when they were added to the database. This is key for differentiating whether or not data should already exist in Postgres.
2. The data becomes slowly available over time. I was hoping by using the word "streaming" a user could make the inference that the Postgres replicator must exist as a long-running process, rather than a run-once and you're done kind of thing.

But like you said, it's all just data in BigQuery so maybe the word should be removed.

cmd/postgres_replicator/README.md
```diff
@@ -0,0 +1,66 @@
...
+        dateField: timeCreated
+      - name: tableExample2
+        dateField: timeCreated
+  - name: datasetExampleName2
+    tables:
+      - name: tableExample3
+        dateField: timeCreated
+```
+
+`BQ_PROJECT_ID`: The GCP project ID where the BigQuery instance resides.
```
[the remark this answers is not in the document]
brandonpaiz: The project name is some friendly title like "My Project", which doesn't have to be unique.
Project ID (the one that should be entered) is some lowercase and uniquiified version of the project name, like `my-project-1234`, and is available on the homepage of a GCP project. Then there is the project number which is just some integer.

I've updated the README to include a little note about how to find the project ID

cmd/postgres_replicator/README.md
brandonpaiz: Added a "how it works" section to the README. I'll add a section to the design doc about libraries I used (bigquery, sql, sqlf)

postgres_replicator/transfer.go
brandonpaiz: Structs, arrays, etc are converted to JSON, which can then be queried / typecast from Postgres queries. I added a section to the README that explains this.

cmd/postgres_replicator/README.md
brandonpaiz: These credentials are used in addition to IAM permissions. I think the cloud proxy is intended to be used when you need to connect to the database locally (outside of some compute environment in GCP). When running on App Engine, IAM authentication is done automatically. (“By default, your app will authorize your connections using an App Engine service account.” https://cloud.google.com/sql/docs/postgres/connect-app-engine-flexible).

postgres_replicator/transfer.go
brandonpaiz: I was worried about this too. It seems like the Golang BigQuery library will [automatically page for you](https://cloud.google.com/bigquery/docs/paging-results)
> The Cloud Client Libraries for Go automatically paginates by default, so you do not need to implement pagination yourself

Just to be sure, I used [this method](https://stackoverflow.com/a/774601) to measure peak memory usage of the replicator when running in App Engine. When transferring a ~9GB table, peak memory usage over several samples was reported to be about 100MB.

I'll add a comment to the function

postgres_replicator/transfer.go
brandonpaiz: I need to build SQL queries programatically based on arbitrary table schema. More popular SQL builder libraries seemed to lack this feature. I've also added this information to the design document.

dashboard/Makefile
```diff
@@ -0,0 +1,24 @@
+GOCMD ?= go
+GOARGS= -trimpath
+REPLICATOR_OUTPUT_DIR ?= bin
+
+replicator: fmt vet
+	$(GOCMD) build $(GOARGS) -o $(REPLICATOR_OUTPUT_DIR)/replicator cmd/postgres_replicator/main.go
+
+configure_replicator:
+	scripts/check_env.sh GCP_PROJECT_ID GCP_DATA_TRANSFER_SERVICE BQ_PROJECT_ID PG_USER PG_PASS PG_DATABASE CLOUD_SQL_INSTANCE
+	cp ../config/dashboard/postgres_replicator/config.yaml postgres_replicator
```
paulosjca, line 10: I don't think you should copy and then substitute in place. It would make more sense to have a generic "envsubst.sh"  that takes one input and one output and performs the substitutions. This can perhaps be done in a follow-up.

dashboard/Makefile
```diff
@@ -0,0 +1,24 @@
+GOCMD ?= go
+GOARGS= -trimpath
+REPLICATOR_OUTPUT_DIR ?= bin
+
+replicator: fmt vet
```
paulosjca, line 5: 1. Make targets have dashes by convention, so "configure-replicator", "configure-grafana".
2. I think it would make sense to move the default configuration into the dashboard directory structure, so config/dashboard -> dashboard/config/default. Under that, dashboards -> grafana to make the copying more logical.
3. We should have environment variables to specify config template locations. CONFIG_TEMPLATE_DIR ?= dashboard/config/default would be a possibility, or specific ones for grafana and the replicator.
4. I would reorder the targets to put grafana first and replicator second (coincidentally, that is alphabetical order).

dashboard/postgres_replicator/logger.go
```diff
@@ -0,0 +1,26 @@
...
+
+// Logger proves per-table logging.
+type Logger struct {
+	prefix string
+	*log.Logger
+}
+
+// NewLogger returns a new Logger.
+func NewLogger(tableName string) *Logger {
+	prefix := "[" + tableName + "]"
```
paulosjca, line 16: I don't think there is a reason to save the prefix in the logger, and to change it during operation. You can always print with format + "[ERROR] " to produce the error string you are trying to produce here.

The prefix you have now is also illogical, since you always add " " to it. You should have fmt.Sprintf("[%s] ", tableName) and use that as the prefix.

dashboard/postgres_replicator/config.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, fmt.Errorf("validation error: %s", err)
+	}
+	overwriteEnvVars(yConfig)
+
+	config := &Config{yConfig}
+	return config, nil
+}
+
+func overwriteEnvVars(conf *YAMLConfig) {
+	postgresPass := os.Getenv("PG_PASS")
```
paulosjca, line 34: Interesting that this variable is the only one overridden here.

dashboard/postgres_replicator/config.go
```diff
@@ -0,0 +1,107 @@
...
+// instance.
+type PostgresConfig struct {
+	DbHost string `yaml:"dbHost"`
+	DbPort string `yaml:"dbPort"`
+	DbUser string `yaml:"dbUser"`
+	DbPass string `yaml:"dbPass"`
+	DbName string `yaml:"dbName"`
+}
+
+// TableConfig stores configuartion about which BigQuery datasets and tables
```
paulosjca, line 97: "configuration". There is a spelling checker for these cases.

dashboard/postgres_replicator/bigquery.go
```diff
@@ -0,0 +1,103 @@
...
+	"strings"
+
+	"cloud.google.com/go/bigquery"
+	"github.com/leporo/sqlf"
+	"google.golang.org/api/iterator"
+)
+
+// BigQueryClient interacts with an instance of BigQuery.
+type BigQueryClient struct {
+	bqClient *bigquery.Client
```
paulosjca, line 16: We should consider what would be needed to test this code in general.

dashboard/Makefile
paulosjca: Since we have a file that comes from a deployment-specific configuration (config.yaml) and one that is generic (app.yaml), we could perhaps organize the configurations like this:

dashboard/config/postgres_replicator/app.yaml
dashboard/config/postgres_replicator/default/config.yaml

Similarly we can have

dashboard/config/grafana/app.yaml ...
dashboard/config/grafana/dashboards/default/...

The common parts can be copied from a fixed location onto the destination by make, with whatever substitutions are required. The "specific" are then copied from the location given by environment variables like CONFIG_TEMPLATE_DIR.

We should also consider whether the postgres replicator needs any substitution at all. The substitutions could all be made by the go code using the environment variables available when the code is deployed.

dashboard/Makefile
brandonpaiz: Done. Now I just use plain old `envsubst` in the make file, taking the files from `dasboard/config` and placing them in a new `containers` directory

dashboard/Makefile
brandonpaiz: 1. Changed underscores to dashes in the Makefile
2. Configuration has been moved into `dashboard/config` as discussed
3. The environment variable `CONFIG_TEMPLATE_DIR` points to the configuration directory and is used in the Makefile
4. Targets reordered

dashboard/postgres_replicator/logger.go
brandonpaiz: The reason for this weird logic was this.
1. Normal logs would have the prefix `[tableName]_` (underscore where the space would be)
2. For error messages, I would put the error tag after the table name, like this: `[tableName][ERROR]_`, and so the location of the space would change to after the error tag.
3. Then the normal tag would need to be restored: `[tableName]_`

The space after the tags had to be moved around. But it's simpler to add the error tag before the table name so I just changed it to do that instead. It removes all that weird space stuff. Also the Logger no longer stores the prefix; I didn't realize there was a built in function.

dashboard/postgres_replicator/config.go
brandonpaiz: Done

dashboard/postgres_replicator/logger.go
```diff
@@ -0,0 +1,27 @@
...
+
+// NewLogger returns a new Logger.
+func NewLogger(tableName string) *Logger {
+	prefix := fmt.Sprintf("[%s] ", tableName)
+	prefixedLogger := log.New(os.Stderr, prefix, log.Ldate|log.Ltime|log.Lmsgprefix)
+	return &Logger{prefixedLogger}
+}
+
+// Errorf adds an "[ERROR]" tag to the log.
+func (tl *Logger) Errorf(format string, v ...interface{}) {
```
paulosjca, line 22: I don't think there is a need to change the prefix here. You could instead change the format to 'Error: ' + format for errors.

brandonpaiz: Fixed :+1:

dashboard/postgres_replicator/logger_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"bytes"
+	"fmt"
+	"log"
+	"testing"
+)
+
+func TestNewLogger(t *testing.T) {
+	tableName := "foobar"
+	logger := NewLogger(tableName)
+	expectedPrefix := fmt.Sprintf("[%s] ", tableName)
```
paulosjca, line 13: Tests should typically not have logic to calculate expected values, because that could replicate the action of the original code. I would also use "got" and "want" for expected and actual.

You may find this package useful in general: https://pkg.go.dev/github.com/google/go-cmp/cmp

dashboard/postgres_replicator/logger_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"testing"
+)
+
+func TestNewLogger(t *testing.T) {
+	tableName := "foobar"
```
paulosjca, line 11: Generally I would also use names like "testTable" instead of using random-looking strings like "foobar".

dashboard/cloudbuild.yaml
```diff
@@ -0,0 +1,37 @@
+steps:
+- name: 'debian:buster'
+  entrypoint: /bin/bash
+  args:
+    - -c
+    - |
+      apt-get update
+      apt-get install -y gettext-base make
+      pushd dashboard
+      make configure_grafana
```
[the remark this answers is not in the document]
brandonpaiz: Done

dashboard/postgres_replicator/logger_test.go
brandonpaiz: Changed!

dashboard/postgres_replicator/logger_test.go
brandonpaiz: Changed!

cmd/postgres_replicator/README.md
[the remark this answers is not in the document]
jtattermusch: IMHO "streaming" is usually used to describe the situation when you're getting realtime updates with new data. This is not our case so let's use different wording.

dashboard/cmd/postgres_replicator/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Automated table creation and type conversion
+
+For non-nested columns, the replicator currently supports
+
+- FLOAT
+- STRING
+- TIMESTAMP
+
+BigQuery table fields that are of the `RECORD`/`REAPEATED`/`STRUCT`/`ARRAY`
```
jtattermusch, line 94: typo: REAPEATED

dashboard/cmd/postgres_replicator/README.md
```diff
@@ -0,0 +1,103 @@
...
+With the latest timestamp, the replicator will query the BigQuery table for rows
+with a newer timestamp. If no timestamp was found (it may be the table was just
+created in Postgres), the replicator will copy all data from the associated
+BigQuery table.
+
+### Automated table creation and type conversion
+
+For non-nested columns, the replicator currently supports
+
+- FLOAT
```
jtattermusch, line 90: qq: we don't need int support because currently the benchmark results schema doesn't contain any ints?

jtattermusch (approved):
Left a few comments, but overall LGTM once @paulosjca  also approves.

dashboard/cmd/postgres_replicator/README.md
brandonpaiz: Fixed, thanks.

dashboard/cmd/postgres_replicator/README.md
brandonpaiz: The benchmarks results schema does have ints, but they're all nested in records so this is fine for now. I have an in-progress fix that will add support for ints and a few other data types.

paulosjca (approved):
Let's address any issues after the initial version is merged and deployed.

PR #245, opened 2021-12-11
Marshal and Unmarshal config file of snapshot of xds server

wanlin31:


containers/runtime/xds/config/config.go
```diff
@@ -0,0 +1,436 @@
...
+	// gRPC listener name must match match the server_target_string in xds:///server_target_string"
+	listenerType := cache.GetResponseType(resource.ListenerType)
+	if _, ok := snap.Resources[listenerType].Items[t.TestGrpcListenerName]; !ok {
+		log.Fatalf("failed validation of listener resource, please set up gRPC listener with name %v", t.TestGrpcListenerName)
+	}
+
+	// Envoy listener's port value is used by routing traffic to Envoy sidecar
+	for listenerName, listenerWithTTL := range snap.Resources[listenerType].Items {
+		// listener port is not required for gRPC listener.
+		if listenerName != t.TestGrpcListenerName {
```
gnossen, line 289: I'm not sure what you mean by "gRPC listener" here.  Do you mean `api_listener`s? If so, this check should probably be made more general so that all `api_listener`s are excluded.

containers/runtime/xds/config/config.go
```diff
@@ -0,0 +1,436 @@
...
+			if forValidation.Address.GetSocketAddress().GetPortValue() != t.TestListenerPort {
+				log.Fatalf("fail to validate Envoy listener's port value: Envoy listener's port value: %v does not match the port that the client target port %v, \n", forValidation.Address.GetSocketAddress().GetPortValue(), t.TestListenerPort)
+			}
+		}
+	}
+
+	// Envoy's dynamic bootstrap file service cluster listed to configure Envoy obtaining configuration from xds server, this cluster name should be listed as the config resource
+
+	// check consistency
+	if err := snap.Consistent(); err != nil {
```
gnossen, line 307: Very cool check. I wasn't aware of this method.

containers/runtime/xds/config/config.go
```diff
@@ -0,0 +1,436 @@
...
+
+// UnmarshalJSON is custom UnmarshalJSON() for CustomSnapshot struct
+func (cs *customSnapshot) UnmarshalJSON(data []byte) error {
+	var values map[string]json.RawMessage
+	json.Unmarshal(data, &values)
+
+	// unmarshal VersionMap
+	versionMap := make(map[string]map[string]string)
+	if err := json.Unmarshal(values["VersionMap"], &versionMap); err != nil {
+		log.Fatalf("TODO: error message %v", err)
```
gnossen, line 78: TODO: error message

containers/runtime/xds/config/config.go
```diff
@@ -0,0 +1,436 @@
...
+	var a anypb.Any
+	if err := protojson.Unmarshal(data, &a); err != nil {
+		log.Fatalf("fail to unmarshal proto.any message: %v", err)
+	}
+
+	switch a.TypeUrl {
+	case resource.EndpointType:
+		parsedEndpoint := endpoint.ClusterLoadAssignment{}
+		if err := anypb.UnmarshalTo(&a, &parsedEndpoint, proto.UnmarshalOptions{}); err != nil {
+			log.Fatalf("fail to unmarshal %v resource: %v", resource.EndpointType, err)
```
gnossen, line 169: Nit: `s/fail/failed/g` throughout in error messages.

gnossen (approved):
Sorry for the wait on the review. Overall, LGTM. The only real issue I see is that we need a little bit of documentation on the config format.

containers/runtime/xds/config/default_config.json
```diff
@@ -0,0 +1,217 @@
...
+                        ]
+                     }
+                  }
+               },
+               "TTL": null
+            }
+         }
+      },
+      {
+         "Version": "",
```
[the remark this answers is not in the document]
wanlin31: There are 8 type of resources: Endpoint, Cluster, Route, ScopedRoute, Listener, Secret, Runtime, ExtensionConfig. In this default config, we have Endpoint, Cluster, Route and Listener, the empty fields are resource of Secret, Runtime, ScopedRoute and ExtensionConfig. Although we don't include these in the default config, the marshaller and unmarshaller have the ability to fill in these resource if the user did supply them, test will also cover these resources.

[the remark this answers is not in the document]
wanlin31: For VersionMap field, it is used internally. not really part of the config I believe.
https://pkg.go.dev/github.com/envoyproxy/go-control-plane@v0.10.0/pkg/cache/v3#Snapshot

containers/runtime/xds/config/default_config.json
```diff
@@ -0,0 +1,217 @@
+{
+   "Resources": [
```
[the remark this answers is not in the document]
wanlin31: I was able to start xDS server using the resource unmarshalled from this default config file, passing validation. I have marshaled the original hardcoded resource (proved to work with both Envoy and gRPC proxyless) into JSON and validate they are exactly the same (except some port number) with this default config. Let me run the xDS server to verify and further update.

containers/runtime/xds/config/default_config.json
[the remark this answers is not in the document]
gnossen: What would happen if these empty objects were removed from this file? Would it still work correctly?

containers/runtime/xds/config/config.go
```diff
@@ -0,0 +1,436 @@
...
+	return nil
+}
+
+// GenerateSnapshotFromConfigFiles takes a default configuration file
+// and user supplied configuration to generate a snapshot
+func (t *TestResource) GenerateSnapshotFromConfigFiles(defaultConfigPath string, userSuppliedConfigPath string) (cache.Snapshot, error) {
+	// read and unmarshal default configuration
+	defaultConfigData, err := os.ReadFile(defaultConfigPath)
+	if err != nil {
+		log.Fatalf("fail to read the default configuration from path (%v): %v", defaultConfigPath, err)
```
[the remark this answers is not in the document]
wanlin31: Thank you soo much! I did not realize I have used fail and failed interchangeably so badly. Fixed

containers/runtime/xds/config/config.go
wanlin31: You are right, the gRPC listeners is just a generic api_listener. Seems like the api_listener is only used for non-proxy listeners. Update it to be more generic. 
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto

containers/runtime/xds/config/config.go
wanlin31: Fixed, thank you.

containers/runtime/xds/config/config.go
wanlin31: Even we skipped port checking for the api_listener, we still need to add the port field in the default config to pass the listener validation check, it should not affect the api_listener, but will double check to verify.

https://source.corp.google.com/piper///depot/google3/third_party/envoy/src/test/server/listener_manager_impl_test.cc;l=4762?q=api_listener

containers/runtime/xds/config/default_config.json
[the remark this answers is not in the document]
wanlin31: I have update the code to make sure they work correctly with or without submitting the placeholders.

containers/runtime/xds/config/config.go
gnossen: > Seems like the api_listener is only used for non-proxy listeners
Actually, the opposite is true. `api_listeners` are queried by proxyless clients.

> Even we skipped port checking for the api_listener, we still need to add the port field in the default config to pass the listener validation check, it should not affect the api_listener, but will double check to verify.

SGTM

containers/runtime/xds/config/default_config.json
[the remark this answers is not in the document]
gnossen: Gotcha. It's just kind of confusing that there are empty objects here without any indication of what they do. It would be nice if we could add comments to this to explain.

Is it possible to add the `@type` field without filling in the rest of the object? If so, let's do that. If not, this is fine as it is.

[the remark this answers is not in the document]
wanlin31: Thank you so much for pointing this out. I have eliminate the empty objects in the default config. :)

containers/runtime/xds/config/default_config.json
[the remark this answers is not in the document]
wanlin31: I have verified that the resource unmarshalled from the configuration JSON file works for both Envoy and gRPC client.

containers/runtime/xds/config/default_config.json
```diff
@@ -0,0 +1,217 @@
+{
```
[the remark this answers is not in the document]
wanlin31: I have wrote some documentations and provide references, let me know if I should get into more details. Thanks!

containers/runtime/xds/config/README.md
```diff
@@ -0,0 +1,33 @@
+# Configuration for xDS server
+
+Configuration JSON files are used to generate the resource snapshot for xDS server. The xDS server consumes these configuration and server these configuration to xDS client or Envoy sidecar.
```
gnossen, line 3: I'd suggest "xDS configuration" instead of "resource snapshot". "Resource snapshot" is sort of an internal name to `go-control-plane`.

containers/runtime/xds/config/README.md
```diff
@@ -0,0 +1,33 @@
...
+* Runtime: [service.runtime.v3.Runtime proto](https://github.com/envoyproxy/envoy/blob/3865cbaec08c5ce44b439eb77e46fe866e37e81a/api/envoy/service/runtime/v3/rtds.proto#L50)
+
+## Default configuration JSON file
+
+The daufult_config.json file contains a piece of configuration that have two listeners with name: `default_testEnvoyListenerName` and `default_testGrpcListenerName`, one Cluster resource, one Route resource and one Endpoint resource. The two listeners are pointing to the same Cluster, eventually the same Endpoint resource.
+
+## User supplied configuration JSON file
+
+If user wish to alter the default configuration, a user defined configuration can be created and once the path of the user supplied configuration file is provided, the xDS server then server the user defined configuration. User supplied configurations are updated on top of the default configuration, so U=user only need to supply the part that wish to alter. If the user did not supply any configuration, the default configuration will be used for xDS server.
```
gnossen, line 33: "U=user"?

gnossen, line 33: Can you please document how the user can supply their own configuration? Is there a fixed path? CLI flag? Environment variable?

gnossen (approved):
This is exactly the sort of documentation I was hoping for! Just a few nits.

containers/runtime/xds/config/README.md
wanlin31: Sorry, typo :)

wanlin31: Actually there is not much to do to use a user defined configuration, the only thing is to create a configuration JSON file just like the default configuration in the config directory. At the time starting an xDS server, there is a flag that can be used to supply the path of the user defined configuration file. 
https://github.com/grpc/test-infra/blob/51516f03ee316e50454a630558c9d2adc5fe11bd/containers/runtime/xds/cmd/main.go#L41

It is a part of the https://github.com/grpc/test-infra/pull/235, I am not so sure about the flag name yet, I can mentioned it here and will update that if needed. 

I will also come up with more explanation on how to use flags in xDS server's README.

gnossen: > create a configuration JSON file just like the default configuration in the config directory

Right. So where is the config directory? In general, what I'm trying to get at is that a newcomer should be able to read the README and know everything they need to supply a custom xDS configuration.

> It is a part of the #235, I am not so sure about the flag name yet, I can mentioned it here and will update that if needed.
Sure. If the feature isn't completed by this PR, we don't need to hold it up for this. We just need to make sure that the README gets to that state at _some_ point.

gnossen (approved):

containers/runtime/xds/config/README.md
```diff
@@ -0,0 +1,37 @@
+# Configuration for xDS server
```
paulosjca, line 1: Would you mind running `prettier --write --print-width=80 --prose-wrap=always` on documentation files?

containers/runtime/xds/config/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Configuration JSON files are used to generate the xDS configuration for xDS server. The xDS server consumes these configuration and server these configuration to xDS client or Envoy sidecar.
```
paulosjca, line 3: xDS server -> an xDS server. these -> this. server -> serves. xDS client -> an xDS client.

containers/runtime/xds/config/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+The configuration JSON files are directly unmarshalled to [Snapshot](https://pkg.go.dev/github.com/envoyproxy/go-control-plane@v0.10.0/pkg/cache/v3#Snapshot) struct.
+
+The configuration files should follow the structure in the daufult_config.json file.
```
paulosjca, line 7: default_config.json. I would also add backticks, `default_config.json`.

containers/runtime/xds/config/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+* Resources: An array holding the resouces, there are currently 8 types of resources, the index of this array corresponds to the resource type.
+* Version: Versions indicate the version of current group of resource. This field is served as key in VersionMap.
+* TTL: Optional fields to set TTL for each resource item.
+* Items: Items are maps within each resource type, the maps' keys are the individual resource names and the values are the actual resources with an optional TTL
+* VersionMap: VersionMap holds the current hash map of all resources in the snapshot. This field should remain nil until it is used. In our use case, we only unmarshal the configuration into Snapshot, the VersionMap remains nil. VersionMap is only to be used with delta xDS.
+
+The source of truth for the fields within each resource in the resource map are their protos.
```
paulosjca, line 15: are -> is.

containers/runtime/xds/config/README.md
```diff
@@ -0,0 +1,37 @@
...
+* ScopedRoute: [config.route.v3.ScopedRouteConfiguration proto](https://github.com/envoyproxy/envoy/blob/3865cbaec08c5ce44b439eb77e46fe866e37e81a/api/envoy/config/route/v3/scoped_route.proto#L83)
+* ExtensionConfig: [config.core.v3.ExtensionConfigSource proto](https://github.com/envoyproxy/envoy/blob/3865cbaec08c5ce44b439eb77e46fe866e37e81a/api/envoy/config/core/v3/extension.proto#L47)
+* Secret: [extensions.transport_sockets.tls.v3.Secret proto](https://github.com/envoyproxy/envoy/blob/3865cbaec08c5ce44b439eb77e46fe866e37e81a/api/envoy/extensions/transport_sockets/tls/v3/secret.proto#L40)
+* Runtime: [service.runtime.v3.Runtime proto](https://github.com/envoyproxy/envoy/blob/3865cbaec08c5ce44b439eb77e46fe866e37e81a/api/envoy/service/runtime/v3/rtds.proto#L50)
+
+## Default configuration JSON file
+
+The daufult_config.json file contains a piece of configuration that have two listeners with name: `default_testEnvoyListenerName` and `default_testGrpcListenerName`, one Cluster resource, one Route resource and one Endpoint resource. The two listeners are pointing to the same Cluster, eventually the same Endpoint resource.
```
paulosjca, line 29: `default_config.json`. have -> has, "with name:" -> named. I wonder if it is necessary to capitalize 'Cluster', 'Route' etc and to say 'resource' every time, or just lowercase and without 'resource' everywhere would be clearer.

paulosjca (approved):
LGTM except for comments on documentation.

containers/runtime/xds/config/README.md
wanlin31: I will keep in mind also in the future, thank you!

containers/runtime/xds/config/README.md
wanlin31: I think what you recommend works. Updated

wanlin31:
This PR create a config library to marshal and unmarshal configuration for xDS server. This PR include:
* The config library provides:
  * Marshal and unmarshal configuration json file for xDS server.
  * Checks to make sure that:
    * The resource are consistent to be served. 
    * Validate that at least one of the listener has the required name
    * Validate at least one of the listener have the proper port.
  * A piece of default configuration json file.
  * Function to update the endpoints. Will report an error if the actual number of the backend are different from the endpoint configurations provided.
* Preliminary documentation for using config library.

PR #255, opened 2022-01-17
Integrate xds server and sidecar container option into existing test infrastructure.

wanlin31:


containers/init/ready/ready.go
```diff
@@ -252,6 +257,50 @@ func WaitForReadyPods(ctx context.Context, ltg LoadTestGetter, pl PodLister, tes
 	return podAddresses, &nodesInfo, nil
 }
 
+func passTarget(quitServer bool, clientIP string, targets []*pb.Endpoint, testType string) error {
```
gnossen, line 260: I don't see any place where `quitServer` is supplied with a value of `false`. Any reason to keep it as a parameter?

containers/init/ready/ready.go
```diff
@@ -252,6 +257,50 @@ func WaitForReadyPods(ctx context.Context, ltg LoadTestGetter, pl PodLister, tes
...
+		log.Fatalf("could not greet: %v", err)
+		statusCode, _ := grpcstatus.FromError(err)
+		log.Print(statusCode.Details()...)
+	}
+	// Used separate goroutines to stop server, non-blocking
+	if quitServer {
+		go func() {
+			fmt.Println("stopping endpoint update server")
+			if _, err := c.QuitTestUpdateServer(ctx, &pb.Void{}); err != nil {
+				statusCode, _ := grpcstatus.FromError(err)
```
gnossen, line 284: How critical is it if this RPC fails?

containers/runtime/xds/README.md
```diff
@@ -0,0 +1,124 @@
...
+  configuration files, the flag here is to make sure that at lease one of the
+  listener resource has the required name.
+
+- -sidecar-listener-port
+
+  This field in only for validation. The listener port that sidecar proxy is
+  listening on. The traffic wish to go through envoy should be send to this
+  port.
+
+- -validation-only
```
gnossen, line 89: Suggestion: Maybe call this `-validate-only` instead? Please also add a sentence clarifying that the server will not actually run if this flag is specified. That wasn't clear to me until I read the code.

containers/runtime/xds/README.md
```diff
@@ -0,0 +1,124 @@
...
+  The path of default configuration JSON file that the resource snapshot based
+  on. The default value of the field is `config/default_config.json`.
+
+- -custom-config-path
+
+  The path of the user supplied configuration JSON file. No default value for
+  this field, for more information check section:
+  [Custom configuration of xDS server](#Custom-configuration-of-xDS-server)
+
+- -non-proxied-target-string
```
gnossen, line 74: I like the extra validation. 👍

containers/runtime/xds/cmd/main.go
```diff
@@ -0,0 +1,135 @@
...
+		l.Errorf("fail to validate the generated snapshot for xDS server: %v", err)
+	}
+
+	l.Infof("xDS server resource snapshot is generated successfully")
+
+	if validationOnly {
+		return
+	}
+	// Move the bootstrap file
+	if pathToBootstrap != "" {
```
gnossen, line 83: Maybe add a comment about _why_ this file is being moved to where it is.

containers/runtime/xds/config/config.go
```diff
@@ -459,3 +453,35 @@ func (t *TestResource) UpdateEndpoint(snap cache.Snapshot) error {
 	}
 	return nil
 }
+
+// SocketListenerOnly takes a pointer of a snapshot, and returns only the socket listeners.
+// This function is used for Proxied test since api_listneners which are used for the
+// non-proxed test can not be validated by Envoy causing the entire resources slices not to be
+// registed.
+func SocketListenerOnly(snap *cache.Snapshot) error {
```
gnossen, line 461: Nit: Function names should be verb phrases.

containers/runtime/xds/logger.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// Debugf print out debug information.
+func (logger Logger) Debugf(format string, args ...interface{}) {
+	log.Printf(format+"\n", args...)
+}
+
+// Infof print out useful information.
+func (logger Logger) Infof(format string, args ...interface{}) {
+	log.Printf(format+"\n", args...)
```
gnossen, line 18: Will this print a timestamp?

kubehelpers/configMap.go
```diff
@@ -0,0 +1,49 @@
...
+package kubehelpers
+
+import (
+	"encoding/json"
+)
+
+// UpdateConfigMapWithServerPort accepts a server port string and a scenarioString string.
+// It returns a updated scenarioString with server port inserted. Currently only supports
+// single scenario.
+// TODO: wanlin31 to update the function to support array of scenarios.
```
gnossen, line 26: Missed TODO?

kubehelpers/psm.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"fmt"
+
+	grpcv1 "github.com/grpc/test-infra/api/v1"
+)
+
+// IsPSMTest checks if a given Loadtest is a PSM test. All clients running
+// a scenario should have the same configuration otherwise a error will be
+// returned.
+func IsPSMTest(clients *[]grpcv1.Client) (bool, error) {
```
gnossen, line 28: It kind of feels like this function is doing two things: validating the test spec and then seeing what type of test spec it is.

Maybe split this out into two functions? Then you would only need one place in the code asserting that all clients must run the same kind of test.

proto/testupdater/test_updater.proto
```diff
@@ -16,16 +16,28 @@ syntax = "proto3";
...
-package endpointupdater;
+package testupdater;
 
-service EndpointUpdater {
+service TestUpdater {
   // Sends an update
-  rpc UpdateEndpoint (EndpointUpdaterRequest) returns (EndpointUpdaterReply) {}
+  rpc UpdateTest (TestUpdateRequest) returns (TestUpdateReply) {}
+
+  rpc QuitTestUpdateServer(Void) returns (Void);
```
gnossen, line 25: Optional: RPCs like this at Google are traditionally called `QuitQuitQuit`

status/status.go
```diff
@@ -54,7 +54,7 @@ func StateForContainerStatus(status *corev1.ContainerStatus) (State, *int32) {
 	if terminateState := status.State.Terminated; terminateState != nil {
 		var state State = Errored
 
-		if terminateState.ExitCode == 0 {
+		if terminateState.ExitCode == 0 || terminateState.ExitCode == 2 {
```
gnossen, line 57: This seems odd. Why is this necessary?

gnossen (commented):
Tremendous progress!

config/constants.go
```diff
@@ -66,10 +66,40 @@ const (
 	// DriverPortEnv specifies the name of the env variable that contains driver port.
 	DriverPortEnv = "DRIVER_PORT"
 
+	// KillAfterEnv specifies the name of the env variable that sets the allowed
+	// response time for a pod after timeout.
+	KillAfterEnv = "KILL_AFTER"
+
+	// NonProxiedTargetStringEnv specifies the target string that PSM test client send
+	// traffic to.
+	NonProxiedTargetStringEnv = "NON_PROXIED_TARGET_STRING"
```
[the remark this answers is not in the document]
wanlin31: Great suggestion, I will update them wherever applies.

containers/runtime/xds/config/config.go
wanlin31: I make it `IncludeSocketListenerOnly`. Thanks.

containers/runtime/xds/logger.go
wanlin31: I updated

containers/init/ready/ready.go
wanlin31: My understanding is since it was separated to a different go-routine, it would not stop the ready container from moving forward. The endpoint (I should have updated the name, I will fix it) update server in the xds server is also in a separate go-routine, I don't think the failure would do anything. Not able to quit it may just cause it to be continuously running.

kubehelpers/configMap.go
wanlin31: Right. I actually want to discuss with you in this. Do you expect we run multiple scenarios in the same LoadTest? Currently I only support update the server port when there is only one scenario.

proto/testupdater/test_updater.proto
wanlin31: I will update it. :)

containers/runtime/xds/logger.go
wanlin31: It does.

status/status.go
wanlin31: As discussed, I will have a separate PR to address this.

containers/init/ready/ready.go
wanlin31: I forgot to add an input argument here, we can probably ask it to check if there is specific ENV, such as "QUITQUITQUIT", is so we quit the update server. Do you see a need that the user might not want to quit the test update server? If not we can probably remove that parameter and quit the test update server always.

api/v1/loadtest_types.go
```diff
@@ -265,6 +319,14 @@ type Client struct {
 	// +optional
 	Build *Build `json:"build,omitempty"`
 
+	// Sidecar describes the Sidecar container, which is the runtime of
+	// the proxy for the test associated with a proxy.
+	Sidecar *Sidecar `json:"sidecar,omitempty"`
```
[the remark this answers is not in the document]
wanlin31: This is a great point, I always thought we need to wrap the container with something, but seems like all the required fields are in the [corev1.Container](https://github.com/kubernetes/api/blob/v0.20.2/core/v1/types.go#L2189). Hi @paulosjca, do you remember why we need to wrap the clone, build and run containers instead of use the container directly?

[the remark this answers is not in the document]
wanlin31: As I discussed offline with @paulosjca, we may change the loadTest configuration definition to be more generic (make all run) before this #255, and the xdsServer and sidecar maybe just Containers.

containers/init/ready/ready.go
gnossen: > Do you see a need that the user might not want to quit the test update server?
Actually, I don't understand why the update server needs to have a remote termination feature at all. Can you please help me understand this?

containers/init/ready/ready.go
gnossen: Closing this thread in favor of [this one](https://github.com/grpc/test-infra/pull/255#discussion_r798138526) since they're essentially the same question.

proto/testupdater/test_updater.proto
gnossen: It's not important. Don't worry about the name change.

kubehelpers/configMap.go
gnossen: I think a single scenario at a time will be fine for now.

gnossen (approved):

PR #260, opened 2022-01-21
fix makefile issue

ybbbby:
The makefile was not working due to the wrong reference to `controller-gen` in https://github.com/grpc/test-infra/pull/209.  
This PR uses old way to refer `controller-gen`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -218,25 +218,32 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
-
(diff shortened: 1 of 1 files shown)


paulosjca:
Thanks for figuring out a fix! It turns out that both controller-gen and kustomize will have issues, because of the deprecation of `go get`. I think the new process of downloading both binaries is a good one, except that it is still using `go get`. I tried with `go install` and it works (except for a known bug in kustomize). PTAL: https://github.com/grpc/test-infra/pull/261.

ybbbby:
> Thanks for figuring out a fix! It turns out that both controller-gen and kustomize will have issues, because of the deprecation of `go get`. I think the new process of downloading both binaries is a good one, except that it is still using `go get`. I tried with `go install` and it works (except for a known bug in kustomize). PTAL: #261.

PR #267, opened 2022-02-03
Update the run containers to be a generic list of containers.

wanlin31:
This commit update the Run field in Driver, Client and Servers to be a generic list of containers. Currently, the first container in the list is considered as the runtime container for the workers, if nothing has been set for the runtime container, the default will be set upon the first container in the list.


wanlin31:
I have include one modified go-test to show how the configuration changes, if the ideas of the PR is okay I will update the rest of the configurations/template.

config/samples/go_example_loadtest.yaml
```diff
@@ -27,6 +27,7 @@ spec:
     language: go
     name: '0'
     run:
+    - name: "worker"
```
paulosjca, line 30: I don't think the examples should have container names, since htey are replaced by default.

podbuilder/podbuilder.go
```diff
@@ -133,7 +134,9 @@ func (pb *PodBuilder) PodForClient(client *grpcv1.Client) (*corev1.Pod, error) {
 	pb.pool = safeStrUnwrap(client.Pool)
 	pb.clone = client.Clone
 	pb.build = client.Build
-	pb.run = &client.Run
+	// Pods are build after setting the default, so at least one container
```
paulosjca, line 137: I would remove these comments.

config/defaults.go
```diff
@@ -150,14 +150,14 @@ func (d *Defaults) setBuildOrDefault(im *imageMap, language string, build *grpcv
 
 // setRunOrDefault sets the default runtime image if it is unset. It returns an
 // error if there is no default runtime image for the provided language.
-func (d *Defaults) setRunOrDefault(im *imageMap, language string, run *grpcv1.Run) error {
-	if run != nil && run.Image == nil {
+func (d *Defaults) setRunOrDefault(im *imageMap, language string, run *corev1.Container) error {
+	if run != nil && run.Image == "" {
```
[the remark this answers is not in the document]
paulosjca: Maybe it should have been empty string from the beginning.

On a related note, perhaps this should take run []*corev1.Container or something like that?

config/samples/go_example_loadtest.yaml
wanlin31: Name is a required field in [Container](https://pkg.go.dev/k8s.io/api/core/v1#Container), it will not pass the validation if we don't have a name. Do we want to always change the first run container's name as run?

config/defaults.go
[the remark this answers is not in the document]
wanlin31: I will make the change to make the function to take run []*corev1.Container.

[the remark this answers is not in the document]
wanlin31: Container.Image is string type, is there any concerns?

config/rbac/role.yaml
```diff
@@ -1,11 +1,13 @@
+
+---
```
wanlin31, line 2: @paulosjca Hi Paulo, I am not sure why it was generated this way, is there any concerns of having this form?

wanlin31:
Once this get merged, I will create a follow up PR to update templates in test-infra and grpc/grpc.

config/constants.go
```diff
@@ -122,6 +123,10 @@ const (
 	// the init containers and containers for a driver or worker pod.
 	WorkspaceVolumeName = "workspace"
 
+	// WorkerContainerName holds the name of the worker container where the test is
```
paulosjca, line 126: Are we still using this constant?

config/constants.go
```diff
@@ -99,7 +99,8 @@ const (
 	RoleLabel = "loadtest-role"
 
 	// RunContainerName holds the name of the main container where the test is
-	// executed.
+	// executed. The runtime for the test may contain multiple run containers.
+	// The main container is always the first container on the list.
 	RunContainerName = "run"
```
paulosjca, line 104: We may want to change this value to "main".

api/v1/loadtest_types.go
```diff
@@ -159,9 +126,9 @@ type Driver struct {
 	// +optional
 	Build *Build `json:"build,omitempty"`
 
-	// Run describes the run container, which is the runtime of the driver for
-	// the actual test.
-	Run Run `json:"run"`
+	// Run describes a list of run container, where the first container is the
```
paulosjca, line 129: I would make sure all three comments to "Run" are the same (plurals and all):

```
// Run describes a list of run containers. The container for the test driver is always
// the first container on the list.
```

I would make the other comments the same, just with client or server instead of driver.

config/constants.go
wanlin31: I thought I have delete it, let me fix it.

config/constants.go
wanlin31: Should we also submit `main` in the load test configuration file for name field for each container?

paulosjca:
I would prefer to see the other PRs as well before this is merged. We will need a new release and deployment before the change on grpc/grpc is merged.

controllers/suite_test.go
```diff
@@ -162,9 +162,9 @@ func newLoadTest() *grpcv1.LoadTest {
 				Name:     optional.StringPtr("driver"),
 				Language: "cxx",
 				Pool:     optional.StringPtr("drivers"),
-				Run: grpcv1.Run{
-					Image: optional.StringPtr("gcr.io/grpc-test-example/driver:v1"),
-				},
+				Run: []corev1.Container{{
+					Image: *optional.StringPtr("gcr.io/grpc-test-example/driver:v1"),
```
paulosjca, line 166: Do we need *optional.StringPtr? This is just a string. Same for the other places where this appears.

config/samples/go_example_loadtest.yaml
```diff
@@ -27,6 +27,7 @@ spec:
     language: go
     name: '0'
     run:
+    - name: "main"
```
paulosjca, line 30: Please make the fields alphabetical.

you should update the templates, run loadtest_examples.sh to generate the examples, then use loadtest_template.py to generate the templates again. Once you do that, the templates should not change (it is a round trip).

I would include all example updates in this PR.

PR #270, opened 2022-02-14
Rename run images flag in configuration generator.

paulosjca:
This includes two changes:

- Rename the flag `-image-prefix` to `-run-image-prefix`, to match `-init-image-prefix` and `build-image-prefix`.
- Pin the latest bazel image version in default configuration to `3.5.0` instead of just `latest`.
- Update Makefile to use RUN_IMAGE_PREFIX and fix format to be more consistent.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/grpc-testing/e2etest/runtime/controller
-  newTag: v0.7.2-kb3.3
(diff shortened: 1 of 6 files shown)


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/grpc-testing/e2etest/runtime/controller
-  newTag: v0.7.2-kb3.3
+  newName: controller
+  newTag: v1.x.x
```
wanlin31, line 16: Can you help me understand what are the use of these two fields?

PR #275, opened 2022-02-16
Integrate of PSM into test-infra

wanlin31:
This commit adds the PSM feature to the test-infra

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@ BUILD_IMAGE_PREFIX ?= ""
 # other than Docker Hub
 RUN_IMAGE_PREFIX ?= ""
 
+# Prefix for all images used for PSM realated tests, enabling use with
+# registries other than DockerHub
+PSM_IMAGE_PREFIX ?= ""
+
 # Image URL to use all building/pushing image targets
 CONTROLLER_IMG ?= $(RUN_IMAGE_PREFIX)controller:$(TEST_INFRA_VERSION)
 
@@ -195,21 +199,21 @@ push-ruby-image: ## Push the Ruby test runtime container image to a registry.
 
 all-psm-images: envoy-image xds-server-image ## Build all psm related container images to a registry.
 
-envoy-image: ## Build the envoy runtime container image.docker build --no-cache -t
-	 ${RUN_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION} containers/runtime/envoy/
+envoy-image: ## Build the envoy runtime container 
+	image.docker build --no-cache -t ${PSM_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION} containers/runtime/envoy/
 
 xds-server-image: ## Build the xds server runtime container image.
-	docker build --no-cache -t ${RUN_IMAGE_PREFIX}xds:${TEST_INFRA_VERSION} -f containers/runtime/xds/Dockerfile .
+	docker build --no-cache -t ${PSM_IMAGE_PREFIX}xds:${TEST_INFRA_VERSION} -f containers/runtime/xds/Dockerfile .
 
 ##@ Publish PSM related container images
 
 push-all-psm-images: push-envoy-image push-xds-server-image ## Push all psm related container images to a registry.
 
 push-envoy-image: ## Push the envoy image to container registry.
-	docker push ${RUN_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION}
+	docker push ${PSM_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION}
 
 push-xds-server-image: ## Push the xds-server image to container registry.
-	docker push ${RUN_IMAGE_PREFIX}xds:${TEST_INFRA_VERSION}
+	docker push ${PSM_IMAGE_PREFIX}xds:${TEST_INFRA_VERSION}
 
 ##@ Deployment
 
--- a/Makefile
+++ b/Makefile
@@ -191,6 +191,26 @@ push-ruby-build-image: ## Push the Ruby build-time container image to a registry
 push-ruby-image: ## Push the Ruby test runtime container image to a registry.
 	docker push $(RUN_IMAGE_PREFIX)ruby:$(TEST_INFRA_VERSION)
 
+##@ Build PSM related container images
+
+all-psm-images: envoy-image xds-server-image ## Build all psm related container images to a registry.
+
+envoy-image: ## Build the envoy runtime container image.docker build --no-cache -t
+	 ${RUN_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION} containers/runtime/envoy/
+
+xds-server-image: ## Build the xds server runtime container image.
+	docker build --no-cache -t ${RUN_IMAGE_PREFIX}xds:${TEST_INFRA_VERSION} -f containers/runtime/xds/Dockerfile .
+
+##@ Publish PSM related container images
+
+push-all-psm-images: push-envoy-image push-xds-server-image ## Push all psm related container images to a registry.
+
+push-envoy-image: ## Push the envoy image to container registry.
+	docker push ${RUN_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION}
+
+push-xds-server-image: ## Push the xds-server image to container registry.
+	docker push ${RUN_IMAGE_PREFIX}xds:${TEST_INFRA_VERSION}
+
 ##@ Deployment
 
 install: install-crd install-rbac ## Install both CRDs and RBACs
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,8 @@ BUILD_IMAGE_PREFIX ?= ""
 # other than Docker Hub
 RUN_IMAGE_PREFIX ?= ""
 
-# Prefix for all images used for PSM realated tests, enabling use with
-# registries other than DockerHub
+# Prefix for all images used for PSM related tests, enabling use with
+# registries other than Docker Hub
 PSM_IMAGE_PREFIX ?= ""
 
 # Image URL to use all building/pushing image targets
@@ -197,21 +197,26 @@ push-ruby-image: ## Push the Ruby test runtime container image to a registry.
 
 ##@ Build PSM related container images
 
+.PHONY: all-psm-images 
 all-psm-images: envoy-image xds-server-image ## Build all psm related container images to a registry.
 
-envoy-image: ## Build the envoy runtime container 
-	image.docker build --no-cache -t ${PSM_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION} containers/runtime/envoy/
+.PHONY: envoy-image 
+envoy-image: ## Build the envoy runtime container image.
+	docker build --no-cache -t ${PSM_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION} containers/runtime/envoy/
 
+.PHONY: xds-server-image
 xds-server-image: ## Build the xds server runtime container image.
 	docker build --no-cache -t ${PSM_IMAGE_PREFIX}xds:${TEST_INFRA_VERSION} -f containers/runtime/xds/Dockerfile .
 
 ##@ Publish PSM related container images
-
+.PHONY: push-all-psm-images
 push-all-psm-images: push-envoy-image push-xds-server-image ## Push all psm related container images to a registry.
 
+.PHONY: push-envoy-image
 push-envoy-image: ## Push the envoy image to container registry.
 	docker push ${PSM_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION}
 
+.PHONY: push-xds-server-image
 push-xds-server-image: ## Push the xds-server image to container registry.
 	docker push ${PSM_IMAGE_PREFIX}xds:${TEST_INFRA_VERSION}
 
--- a/containers/init/ready/ready.go
+++ b/containers/init/ready/ready.go
@@ -429,6 +429,9 @@ func main() {
 		} else {
 			log.Fatalf("failed to obtain PSM server target")
 		}
+
+		log.Println("waiting 20s in case we are running a proxied test where we need Envoy to be fully started")
+		time.Sleep(20 * time.Second)
 	}
 
 	outputMetadataFile := DefaultMetadataOutputFile
--- a/containers/init/ready/ready.go
+++ b/containers/init/ready/ready.go
@@ -289,7 +289,6 @@ func communicateWithEachClient(clientIP string, targets []*pb.Endpoint, isProxie
 			log.Print(statusCode.Details()...)
 		}
 	}()
-
 	return psmServerTargetOverride, nil
 }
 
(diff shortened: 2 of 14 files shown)


Makefile
```diff
@@ -16,6 +16,10 @@ BUILD_IMAGE_PREFIX ?= ""
 # other than DockerHub
 RUN_IMAGE_PREFIX ?= ""
 
+# Prefix for all images used for PSM realated tests, enabling use with
+# registries other than DockerHub
```
paulosjca, line 20: realated -> related. DockerHub -> Docker Hub. I have already made the second change for other occurrences.

kubehelpers/configMap.go
```diff
@@ -0,0 +1,49 @@
+/*
+Copyright 2020 gRPC authors.
```
paulosjca, line 2: New files should have a date of 2022.

kubehelpers/configMap.go
```diff
@@ -0,0 +1,49 @@
...
+package kubehelpers
+
+import (
+	"encoding/json"
+)
+
+// UpdateConfigMapWithServerPort accepts a server port string and a scenarioString string.
+// It returns a updated scenarioString with server port inserted. Currently only supports
+// single scenario.
+// TODO: wanlin31 to update the function to support array of scenarios.
```
paulosjca, line 26: Format should be TODO(wanlin31): update the function to support an array of scenarios.

kubehelpers/configMap.go
```diff
@@ -0,0 +1,49 @@
...
+*/
+
+package kubehelpers
+
+import (
+	"encoding/json"
+)
+
+// UpdateConfigMapWithServerPort accepts a server port string and a scenarioString string.
+// It returns a updated scenarioString with server port inserted. Currently only supports
```
paulosjca, line 24: "returns an updated"

kubehelpers/configMap_test.go
```diff
@@ -0,0 +1,38 @@
+/*
+Copyright 2020 gRPC authors.
```
paulosjca, line 2: 2020 -> 2022.

kubehelpers/psm.go
```diff
@@ -0,0 +1,84 @@
+/*
+Copyright 2020 gRPC authors.
```
paulosjca, line 2: 2020 -> 2022.

kubehelpers/psm.go
```diff
@@ -0,0 +1,84 @@
...
+func IsProxiedTest(clients *[]grpcv1.Client) bool {
+	for _, c := range *clients {
+		if ContainerForName("envoy", c.Run) != nil {
+			return true
+		}
+	}
+	return false
+}
+
+// IsClientsSpecValid checks if the given set of the client specs are valid.
```
paulosjca, line 44: "is valid".

kubehelpers/psm.go
```diff
@@ -0,0 +1,84 @@
...
+func IsPSMTest(clients *[]grpcv1.Client) bool {
+	for _, c := range *clients {
+		if ContainerForName("xds-server", c.Run) != nil {
+			return true
+		}
+	}
+	return false
+}
+
+// IsProxiedTest is to check if the current test have sidecar container
```
paulosjca, line 33: "checks if the current test has a sidecar container specified."

"This check must be performed after validating the client specs."

kubehelpers/psm.go
```diff
@@ -0,0 +1,84 @@
...
+
+package kubehelpers
+
+import (
+	"fmt"
+
+	grpcv1 "github.com/grpc/test-infra/api/v1"
+)
+
+// IsPSMTest checks if a given Loadtest is a PSM test. This check must be
```
paulosjca, line 22: "IsPSMTest checks if a given LoadTest is a (proxied or proxyless) service mesh test."

"This test must be performed after validating the client specs."

kubehelpers/psm.go
```diff
@@ -0,0 +1,84 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// IsClientsSpecValid checks if the given set of the client specs are valid.
+func IsClientsSpecValid(clients *[]grpcv1.Client) (bool, error) {
+	if len(*clients) == 0 {
+		err := fmt.Errorf("no client specified in the given load test")
```
paulosjca, line 47: "no client specified". may be enough.

kubehelpers/psm.go
```diff
@@ -0,0 +1,84 @@
...
+
+	for _, c := range *clients {
+		if ContainerForName("xds-server", c.Run) != nil {
+			numberOfClientWithXdsServer++
+			if ContainerForName("envoy", c.Run) != nil {
+				numberOfClientWithSidecar++
+			}
+		} else {
+			if ContainerForName("envoy", c.Run) != nil {
+				err := fmt.Errorf("encountered a client specified a sidecar container without specifyling xds-server container")
```
paulosjca, line 61: "encountered a client with envoy container but no xds-server container"

kubehelpers/psm.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+
+	if numberOfClientWithXdsServer == 0 {
+		return true, nil
+	}
+
+	if numberOfClientWithSidecar == 0 {
+		if len(*clients) != numberOfClientWithXdsServer {
+			err := fmt.Errorf("only some of the clients have xds-server container specified")
```
paulosjca, line 73: "encountered some clients with xds-server container and some without xds-server container"

kubehelpers/psm.go
```diff
@@ -0,0 +1,84 @@
...
+			return false, err
+		}
+	} else {
+		if len(*clients) != numberOfClientWithSidecar {
+			err := fmt.Errorf("only some of the clients have envoy container specified")
```
paulosjca, line 78: "encountered some clients with envoy container and some without envoy container"

Makefile
```diff
@@ -189,6 +193,26 @@ push-ruby-build-image: ## Push the Ruby build-time container image to a registry
 push-ruby-image: ## Push the Ruby test runtime container image to a registry.
 	docker push $(RUN_IMAGE_PREFIX)ruby:$(TEST_INFRA_VERSION)
 
+##@ Build PSM related container images
+
+all-psm-images: envoy-image xds-server-image ## Build all psm related container images to a registry.
+
+envoy-image: ## Build the envoy runtime container 
+	image.docker build --no-cache -t ${PSM_IMAGE_PREFIX}sidecar:${TEST_INFRA_VERSION} containers/runtime/envoy/
```
gnossen, line 201: Pretty sure `image.docker` is a line rewrap mistake.

Makefile
```diff
@@ -189,6 +193,26 @@ push-ruby-build-image: ## Push the Ruby build-time container image to a registry
 push-ruby-image: ## Push the Ruby test runtime container image to a registry.
 	docker push $(RUN_IMAGE_PREFIX)ruby:$(TEST_INFRA_VERSION)
 
+##@ Build PSM related container images
+
+all-psm-images: envoy-image xds-server-image ## Build all psm related container images to a registry.
```
gnossen, line 198: Nit: All of these targets that don't correspond to a file on the filesystem should probably be [`PHONY`](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html#:~:text=A%20phony%20target%20is%20one,name%2C%20and%20to%20improve%20performance.)

containers/init/ready/ready.go
```diff
@@ -252,6 +261,48 @@ func WaitForReadyPods(ctx context.Context, ltg LoadTestGetter, pl PodLister, tes
 	return podAddresses, &nodesInfo, nil
 }
 
+func communicateWithEachClient(clientIP string, targets []*pb.Endpoint, isProxied bool) (string, error) {
```
gnossen, line 264: Docstring explaining this function and the meaning of each parameter and return value please.

containers/init/ready/ready.go
```diff
@@ -252,6 +261,48 @@ func WaitForReadyPods(ctx context.Context, ltg LoadTestGetter, pl PodLister, tes
...
+	var psmServerTargetOverride string
+	dialTarget := net.JoinHostPort(clientIP, fmt.Sprint(testconfig.ServerUpdatePort))
+	conn, err := grpc.Dial(dialTarget, grpc.WithInsecure())
+	if err != nil {
+		log.Fatalf("did not connect: %v", err)
+	}
+	defer conn.Close()
+	c := pb.NewTestUpdaterClient(conn)
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
gnossen, line 273: Please raise this 20 to a constant at the top of the file so it can be modified more easily.

containers/init/ready/ready.go
```diff
@@ -324,13 +375,91 @@ func main() {
...
+			if err := ioutil.WriteFile(DefaultServerTargetOverrideOutputFile, []byte(psmTargetOverride), 0777); err != nil {
+				log.Fatalf("failed to write PSM server target: %v to %v", psmTargetOverride, DefaultServerTargetOverrideOutputFile)
+			}
+			log.Printf("write PSM server target: %v to %v\n", psmTargetOverride, DefaultServerTargetOverrideOutputFile)
+		} else {
+			log.Fatalf("failed to obtain PSM server target")
+		}
+
+		if isProxied {
+			envoyListenerPort := strings.Split(psmTargetOverride, ":")[1]
```
gnossen, line 430: Is this string validated anywhere? I'm wondering about an out-of-bounds access.

containers/runtime/driver/run.sh
```diff
@@ -19,8 +19,13 @@ if [ -n "${QPS_WORKERS_FILE}" ]; then
   export QPS_WORKERS=$(cat "${QPS_WORKERS_FILE}")
 fi
 
+
+if [ -f /var/data/qps_workers/server_target_override ]; then
```
gnossen, line 23: Why have the conditional if the script will fail anyway if the file is not present?

kubehelpers/configMap.go
```diff
@@ -0,0 +1,49 @@
...
+	var jsonScenarioMap map[string]map[string]json.RawMessage
+	if err := json.Unmarshal([]byte(scenarioString), &jsonScenarioMap); err != nil {
+		return updatedScenarios, err
+	}
+	var serverConfig map[string]interface{}
+	if err := json.Unmarshal(jsonScenarioMap["scenarios"]["server_config"], &serverConfig); err != nil {
+		return updatedScenarios, err
+	}
+	serverConfig["port"] = port
+	serverConfigByte, err := json.Marshal(serverConfig)
```
gnossen, line 38: Nit: `s/serverConfigByte/serverConfigBytes`. Plural, not singular.

kubehelpers/configMap.go
```diff
@@ -0,0 +1,49 @@
...
+	updatedScenarios := ""
+	var jsonScenarioMap map[string]map[string]json.RawMessage
+	if err := json.Unmarshal([]byte(scenarioString), &jsonScenarioMap); err != nil {
+		return updatedScenarios, err
+	}
+	var serverConfig map[string]interface{}
+	if err := json.Unmarshal(jsonScenarioMap["scenarios"]["server_config"], &serverConfig); err != nil {
+		return updatedScenarios, err
+	}
+	serverConfig["port"] = port
```
gnossen, line 37: Is it possible for this field to already be specified somehow? More generally, can you list all of the components that write to this ConfigMap?

kubehelpers/psm.go
```diff
@@ -0,0 +1,84 @@
...
+func IsClientsSpecValid(clients *[]grpcv1.Client) (bool, error) {
+	if len(*clients) == 0 {
+		err := fmt.Errorf("no client specified in the given load test")
+		return false, err
+	}
+	var numberOfClientWithSidecar int
+	var numberOfClientWithXdsServer int
+
+	for _, c := range *clients {
+		if ContainerForName("xds-server", c.Run) != nil {
```
gnossen, line 54: The container names "xds-server" and "envoy" have been used in several places. Please use constants so that these names can be more easily changed.

podbuilder/podbuilder.go
```diff
@@ -152,6 +153,23 @@ func (pb *PodBuilder) PodForClient(client *grpcv1.Client) (*corev1.Pod, error) {
 		Name:  config.DriverPortEnv,
 		Value: fmt.Sprint(config.DriverPort)})
 
+	if xdsServer := kubehelpers.ContainerForName("xds-server", pod.Spec.Containers); xdsServer != nil {
```
gnossen, line 156: Repeating from above, please use a constant for these container names.

podbuilder/podbuilder.go
```diff
@@ -152,6 +153,23 @@ func (pb *PodBuilder) PodForClient(client *grpcv1.Client) (*corev1.Pod, error) {
...
+			pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{Name: "grpc-xds-bootstrap"})
+
+			runContainer.VolumeMounts = append(runContainer.VolumeMounts, corev1.VolumeMount{
+				Name:      "grpc-xds-bootstrap",
+				MountPath: "/bootstrap",
+				ReadOnly:  false,
+			})
+			xdsServer.VolumeMounts = append(xdsServer.VolumeMounts, corev1.VolumeMount{
+				Name:      "grpc-xds-bootstrap",
+				MountPath: "/bootstrap",
```
gnossen, line 167: This is unexpected to me. How is the xDS server using the bootstrap file?

gnossen (approved):
At a high level, it would be nice to see a little bit more explanation of the design behind the change and the rationale behind it in the PR description.

Overall, LGTM.

Makefile
wanlin31: Thank you so much to let me know, I didn't know this before. :)

containers/init/ready/ready.go
wanlin31: Missed it, I was using 20 for testing, forgot to update it. Thank you for pointing out!

containers/init/ready/ready.go
wanlin31: Added a check here.

kubehelpers/psm.go
wanlin31: I will put them in the psm_constant.go.

podbuilder/podbuilder.go
wanlin31: The clients need the bootstrap, however, it would not be very convenient to rebuilt each worker's image to include bootstrap file, better options is to have the bootstrap file in a more centralized image, either the clone image (worker's init container) or xDS server image. 
We used xDS server's image instead of the clone image to minimized the change for a regular test (i.e. for people who doesn't run PSM test, if they see the logic of potentially moving a file in the clone image, they might be confused). Also we always need to have xDS server when running PSM tests, the xDS server image makes a better choice to store the bootstrap file.

Since the bootstrap file is stored inside the xDS image, I need to move the bootstrap file to a shared volume (between xds server and the client) so that the client would have access to it. Let me know if I have not explained it well. :)

containers/runtime/driver/run.sh
wanlin31: I might have understand this wrong. -f is to check if the value of `$(cat /var/data/qps_workers/server_target_override)` can be passed to SERVER_TARGET_OVERRIDE. If I `cat` a non-existing file it would return an error and terminate the process. Regardless of having a value for SERVER_TARGET_OVERRIDE, the test would run. Inside the driver, there is a length check for the target provided through `--qps_server_target_override`, if the length is zero, the driver just ignore this flag. (https://github.com/grpc/grpc/blob/7c74d58cbecc8342c349cbf39eecb9aa0b666522/test/cpp/qps/driver.cc#L482) I might should be more specific to make it less confusing.

podbuilder/podbuilder.go
gnossen: I see. That makes sense. But both of these mounts have `ReadOnly: false`? Based on your explanation above, the `runContainer` only _reads_ the bootstrap file -- it doesn't write to it. So shouldn't `ReadOnly: true` be set?

wanlin31: I can make them read only: true.

wanlin31: Right, I think it is okay to have readonly for run container. Thanks!

kubehelpers/configMap.go
wanlin31: I have add a log where the test server port has been override.

paulosjca (approved):

PR #276, opened 2022-02-16
Reverse order of deploy and install targets.

paulosjca:


wanlin31:
Is this fix the issue found from deploying on a new cluster?

ybbbby:
> Is this fix the issue found from deploying on a new cluster?

Yep:). The `deploy` creates the namespace while `install` would create under that namespace.

ybbbby (approved):

PR #280, opened 2022-02-28
Update directory names.

wanlin31:
The commit updates the envoy to sidecar and xds to xds-server
for all related components to be consistent.


Makefile
```diff
@@ -206,7 +206,7 @@ envoy-image: ## Build the envoy runtime container image.
 
 .PHONY: xds-server-image
 xds-server-image: ## Build the xds server runtime container image.
-	docker build --no-cache -t ${PSM_IMAGE_PREFIX}xds:${TEST_INFRA_VERSION} -f containers/runtime/xds/Dockerfile .
+	docker build --no-cache -t ${PSM_IMAGE_PREFIX}xdsServer:${TEST_INFRA_VERSION} -f containers/runtime/xds/Dockerfile .
```
paulosjca, line 209: I think the image name and the directory in the repository should be `xds-server`, to match the target.

The same is also true for envoy and the envoy-image target, the directory and the target and the image name souhld match and be all `sidecar`.

wanlin31: I have update the PR, thank you.

paulosjca (approved):

PR #285, opened 2022-03-08
Add pre-built image for grpc-dotnet

jtattermusch:
Adds the Dockerfile for "prebuilt" version of grpc-dotnet qps worker.

### DIFF
--- a/containers/pre_built_workers/dotnet/Dockerfile
+++ b/containers/pre_built_workers/dotnet/Dockerfile
@@ -1,0 +1,51 @@
+# Copyright 2022 The gRPC Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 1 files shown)


jtattermusch:
CC @JamesNK

containers/pre_built_workers/dotnet/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN git submodule update --init
+
+# Save commit sha for debug use
+RUN echo 'COMMIT SHA' > GRPC_GIT_COMMIT.txt
+RUN git rev-parse $GITREF >> GRPC_GIT_COMMIT.txt
+
+# Build standalone build of qps worker, put output into qps_worker directory.
+RUN dotnet publish -c Release -o qps_worker perf/benchmarkapps/QpsWorker/
+
+# Note that the libgrpc_csharp_ext.so library built by the build image
```
JamesNK, line 41: Is this comment relevant for grpc-dotnet?

paulosjca (commented):
I think it makes more sense to wait until the examples are generated from templates before checking this in. This change is not required for continuous integration, because continuous integration uses the templates in grpc/grpc, but does not use the examples.

containers/pre_built_workers/dotnet/Dockerfile
jtattermusch: Good point. Updated.

jtattermusch:
I updated the dockerfile according to comments. I tested building the image locally off of the master branch of grpc-dotnet.
Since the dockerfile is required to be able to run the benchmarks on the CI (even for just experiments), I'm going to merge this PR and address TODOs in a followup PR.

TODO: add support for grpc-dotnet in non-pre-built image (not immediately necessary for the CI runs, since they only use pre-built images)

TODO: regenerate the examples based on https://github.com/grpc/grpc/blob/master/tools/run_tests/performance/README.md#generating-load-test-examples

PR #287, opened 2022-03-17
Add complete configuration for prometheus installation.

wanlin31:
This commit add complete configurations for installing prometheus operator and prometheus.


paulosjca (approved):
LGTM with the note that we want to automate this deployment using the Makefile. This can be done in a follow-up PR.

wanlin31:
> LGTM with the note that we want to automate this deployment using the Makefile. This can be done in a follow-up PR.

Right, I have tried to apply all the file together use, did not observe a need for specific order or wait time within them:
```
kubectl create -f config/prometheus/setup/
```

PR #291, opened 2022-04-01
Change driver version

wanlin31:
This is to pin the driver version to include https://github.com/grpc/grpc/pull/29160.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,8 @@
 TEST_INFRA_VERSION ?= latest
 
 # Version of gRPC core used for the gRPC driver
-DRIVER_VERSION ?= v1.45.0
+# Pinned to include https://github.com/grpc/grpc/pull/29160
+DRIVER_VERSION ?= 2fd632a4c1d22abc218cfb0e04f896a1d6081ec1
 
 # Prefix for all images used as clone and ready containers, enabling use with
 # registries other than Docker Hub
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,7 +2,8 @@
 TEST_INFRA_VERSION ?= latest
 
 # Version of gRPC core used for the gRPC driver
-DRIVER_VERSION ?= v1.45.0
+# Pinned to include https://github.com/grpc/grpc/pull/29160
```
ybbbby, line 5: If this is the only pr needed, we probably wan to pin to that particular commit? It is also okay to just keep like this if we would like to update this very soon in the next several PRs.

ybbbby (approved):
LGTM, one NIT

Makefile
wanlin31: I picked a newer commit to include other changes needed for our release 1.2.0, such as https://github.com/grpc/grpc/pull/29207, https://github.com/grpc/grpc/pull/29197 and https://github.com/grpc/grpc/pull/29057. I will update the comment to remove the specific 
reason for pinning the driver, we can also figure something out next Monday.

wanlin31: I have updated the commit to just include what is necessary for v1.2.0

paulosjca: For future reference, if we are making a change, we should explain what we are doing. If we are updating to include three things, we should mention all three things and then choose the newest of the three commits to pin on. Also, one reason why we are doing this is because these changes are necessary to support the new functionality in this release, and grpc/grpc release 1.46.0 is not out yet.

PR #294, opened 2022-04-06
Add grpc-dotnet support (contd.)

jtattermusch:
More progress toward fully supporting grpc-dotnet.

### DIFF
--- a/config/defaults_template.yaml
+++ b/config/defaults_template.yaml
@@ -18,6 +18,10 @@ languages:
   buildImage: "{{ .BuildImagePrefix }}csharp:{{ .Version }}"
   runImage: mcr.microsoft.com/dotnet/runtime:3.1-bullseye-slim
 
+- language: dotnet
+  buildImage: "{{ .BuildImagePrefix }}dotnet:{{ .Version }}"
+  runImage: "{{ .RunImagePrefix }}dotnet:{{ .Version }}"
+
 - language: cxx
   buildImage: l.gcr.io/google/bazel:3.5.0
   runImage: "{{ .RunImagePrefix }}cxx:{{ .Version }}"
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ delete_prebuilt_workers: fmt vet ## Build the delete_prebuilt_workers tool binar
 
 ##@ Build container images
 
-all-images: clone-image controller-image csharp-build-image cxx-image driver-image go-image java-image node-build-image node-image php7-build-image php7-image python-image ready-image ruby-build-image ruby-image ## Build all container images.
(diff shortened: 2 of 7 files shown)


jtattermusch:
While the process for adding a new language to support the "prebuilt" images is straightforward, the process for adding a language to the "basic" workflow is quite painful.  I added the "basic" docker images for grpc-dotnet (build and runtime images) and I managed to generate the example for grpc-dotnet, but it's not clear to me how I should test all this,
since the generated example depends on the default docker image for a given language to be already deployed to a controller (which would require me to actually do a release myself).

- what should be my next step to make sure the build and runtime images I wrote actually work?
- is there a way we could simplify the process for adding a new language for the "basic" workflow ("prebuilt" seems straightforward)?

config/defaults_template.yaml
```diff
@@ -18,6 +18,10 @@ languages:
   buildImage: "{{ .BuildImagePrefix }}csharp:{{ .Version }}"
   runImage: mcr.microsoft.com/dotnet/runtime:3.1-bullseye-slim
 
+- language: dotnet
```
jtattermusch, line 21: I guess I'd have to update some buildfiles as well, but it is not clear which ones.
Anything else that need to be done after adding the "build" and "runtime" images for dotnet?

paulosjca: You should add targets to the top-level Makefile, modeled on csharp-build-images and csharp-push-images.

You can test this by building and deploying the controller to a test cluster, internally we have "benchmarks-dev1" for that.

I will add some documentation on adding a language to make the steps clearer.

jtattermusch:
I did this to test the dotnet runtime and build docker images:

Build the testing version of the docker images
`make TEST_INFRA_VERSION=20220408 BUILD_IMAGE_PREFIX=gcr.io/grpc-testing/e2etest/jtattermusch/build/ RUN_IMAGE_PREFIX=gcr.io/grpc-testing/e2etest/jtattermusch/runtime/ dotnet-image dotnet-build-image push-dotnet-image push-dotnet-build-image`

Locally modified the dotnet example to use those docker images:
e.g.
`image: gcr.io/grpc-testing/e2etest/jtattermusch/runtime/dotnet:20220408`

Ran `kubectl apply -f config/samples/dotnet_example_loadtest.yaml`
- the test has passed.

config/defaults_template.yaml
jtattermusch: updated the makefiles PTAL. Came up with a simpler way to test the docker images for dotnet (see my last comment).

jtattermusch:
Hopefully this is now ready for final review?

paulosjca (approved):

PR #297, opened 2022-04-11
fix postgres replicator logger

brandonpaiz:
Format strings were not printing properly


wanlin31 (approved):

dashboard/postgres_replicator/logger.go
```diff
@@ -20,5 +20,5 @@ func NewLogger(tableName string) *Logger {
 
 // Errorf adds an "Error: " prefix to the format string
 func (l *Logger) Errorf(format string, v ...interface{}) {
-	l.Printf("Error: " + format)
+	l.Printf(fmt.Sprintf("Error: %s", format), v...)
```
wanlin31, line 23: Can I ask why it didn't work before? It seems to me the two approach are equivalent.

PR #306, opened 2022-04-21
Collect logs from all containers and add log to test properties.

wanlin31:
This PR updated the infrastructure to save logs to file from all existing container which produces logs.  We have also separate the property generation of name property and log properties to avoid dependencies.

We also add the log link to the Xunit reports when there are pod logs.

### DIFF
--- a/tools/runner/logsaver.go
+++ b/tools/runner/logsaver.go
@@ -29,26 +29,24 @@ import (
 	corev1types "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
-// SaveAllLogs saves all logs to files and return their LogInfo.
-// This function goes through all the containers in each pod and write
-// the log to the a given path if the container have logs. The name
-// of the files are in format pod-name-container-name.log. After
-// writing the logs the function returns a list of pointers of
-// LogInfo objects containing the log's information.
+// SaveAllLogs saves all container logs to files under a given directory.
+// This function goes through every container in every pods and writes
+// its log to a file, if it is not empty. Information about each saved log
+// is returned as a pointer to a LogInfo object.
 func SaveAllLogs(ctx context.Context, loadTest *grpcv1.LoadTest, podsGetter corev1types.PodsGetter, pods []*corev1.Pod, podLogDir string) ([]*LogInfo, error) {
 	var logInfos []*LogInfo
 
-	// Attempt to create directory. Will not error if directory already exists
+	// Attempts to create directory. Will not error if directory already exists.
 	err := os.MkdirAll(podLogDir, os.ModePerm)
 	if err != nil {
 		return logInfos, fmt.Errorf("failed to create pod log output directory %s: %v", podLogDir, err)
 	}
 
-	// Write logs to files
+	// Write logs to files.
 	for _, pod := range pods {
 		for _, container := range pod.Spec.Containers {
 
-			logInfo, err := SaveLog(ctx, loadTest, pod, podsGetter, container.Name, podLogDir)
+			logInfo, err := SaveLog(ctx, loadTest, podsGetter, pod, container.Name, podLogDir)
 			if err != nil {
 				return logInfos, fmt.Errorf("could not get log from container: %v", err)
 			}
@@ -61,12 +59,11 @@ func SaveAllLogs(ctx context.Context, loadTest *grpcv1.LoadTest, podsGetter core
 	return logInfos, nil
 }
 
-// SaveLog retrieves and save logs for a specific container.
-// This function retrieves logs from a container under given
-// container name within given pod,then save the logs to the
-// given file path, if there are logs to save. The function
-// also returns a point of the LogInfo object.
-func SaveLog(ctx context.Context, loadTest *grpcv1.LoadTest, pod *corev1.Pod, podsGetter corev1types.PodsGetter, containerName string, podLogDir string) (*LogInfo, error) {
+// SaveLog retrieves and saves logs for a specific container.
+// This function retrieves the log for a single container within a given
+// pod, and writes it to a file, if it is not empty. Information about
+// the saved log is returned as a pointer to a LogInfo object.
+func SaveLog(ctx context.Context, loadTest *grpcv1.LoadTest, podsGetter corev1types.PodsGetter, pod *corev1.Pod, containerName string, podLogDir string) (*LogInfo, error) {
 	req := podsGetter.Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{Container: containerName})
 	containerLogs, err := req.Stream(ctx)
 	if err != nil {
@@ -77,12 +74,12 @@ func SaveLog(ctx context.Context, loadTest *grpcv1.LoadTest, pod *corev1.Pod, po
 	logBuffer := new(bytes.Buffer)
 	logBuffer.ReadFrom(containerLogs)
 
-	// Don't write empty buffers
+	// Don't write empty buffers,
 	if logBuffer.Len() == 0 {
 		return nil, nil
 	}
 
-	// Open output file
+	// Open output file,
 	filePath := filepath.Join(podLogDir, LogFileName(pod.Name, containerName))
 	file, err := os.Create(filePath)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


paulosjca (commented):
I think we should collect logs for all containers. Later we should also add links to the logs to the properties stored in the xUnit report.

wanlin31:
> I think we should collect logs for all containers. Later we should also add links to the logs to the properties stored in the xUnit report.

Updated the pr per comment.

Verified by running experimental CI, details can be found in the bug. Also verified that not providing the `log-url-prefix` flag would not result in test failure.

tools/runner/runner.go
```diff
@@ -146,9 +146,16 @@ func (r *Runner) runTest(ctx context.Context, config *grpcv1.LoadTest, reporter
 				reporter.Error("Could not save pod logs: %s", err)
 			}
 			reporter.AddProperty("name", loadTest.Name)
-			for property, value := range savedLogs.GenerateProperties(loadTest) {
+			for property, value := range savedLogs.GenerateNameProperties(loadTest) {
```
paulosjca, line 149: I would leave the "pod." and "log." out of the functions that generate properties, and instead add that here in the loops.

I am actually thinking that we could pass in a []string of prefixes to the property functions and join the properties in a standardized way inside.

tools/runner/properties.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// LogInfo contains infomation for a log entry.
+type LogInfo struct {
+	podNameElement string
```
paulosjca, line 28: Let's have docstrings on each field of the struct. Is there a reason why they are not public?

tools/runner/properties.go
```diff
@@ -0,0 +1,74 @@
...
+package runner
+
+import (
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// LogInfo contains infomation for a log entry.
```
paulosjca, line 26: Each log, each saved log, each log file, but not "log entry" (log entry is a line in a log).

tools/runner/properties.go
```diff
@@ -0,0 +1,74 @@
...
+		url := logURLPrefix + logInfo.logPath
+		properties[podLogPropertyKey] = url
+	}
+	return properties
+}
+
+// PodNameElement trim off the loadtest name from the pod name,
+// and return only the element of pod name such as client-0,
+// driver-0 and server-0.
+func PodNameElement(podName, loadTestName string) string {
```
paulosjca, line 58: Perhaps "Elem" would be more in line with golang than "Element".

tools/runner/runner.go
```diff
@@ -141,14 +145,23 @@ func (r *Runner) runTest(ctx context.Context, config *grpcv1.LoadTest, reporter
 		status = statusString(config)
 		switch {
 		case loadTest.Status.State.IsTerminated():
-			savedLogs, err := r.logSaver.SavePodLogs(ctx, loadTest, outputDir)
+			pods, err := GetTestPods(ctx, loadTest, r.podsGetter)
+			if err != nil {
+				reporter.Error("Could not list all pods belongs to %s: %s", loadTest.Name, err)
+			}
+			savedLogInfos, err := SaveLogs(ctx, loadTest, pods, r.podsGetter, outputDir)
```
paulosjca, line 152: Perhaps the order here should be r.PodsGetter, pods.

tools/runner/properties.go
```diff
@@ -0,0 +1,74 @@
...
+
+// LogInfo contains infomation for a log entry.
+type LogInfo struct {
+	podNameElement string
+	containerName  string
+	logPath        string
+}
+
+// NewLogInfo creates a pointer of new LogInfo object.
+func NewLogInfo(podNameElement string, containerName string, logPath string) *LogInfo {
```
paulosjca, line 34: Is this necessary? You can just create the struct from the fields, there is no function here.

paulosjca (commented):
Remember to update the PR description.

tools/runner/runner.go
```diff
@@ -50,18 +51,21 @@ type Runner struct {
 	// deleteSuccessfulTests determines whether tests that terminate without
 	// errors should be deleted immediately.
 	deleteSuccessfulTests bool
-	// logSaver saves pod log files
-	logSaver *LogSaver
+	// podsLister obtain a list of pods
```
paulosjca, line 54: Perhaps I would add podsGetter right after loadTestGetter.

tools/runner/properties.go
```diff
@@ -0,0 +1,74 @@
...
+		logPath:        logPath,
+	}
+}
+
+// PodLogProperties creates container log property name to
+// container log link map.
+func PodLogProperties(logInfos []*LogInfo, logURLPrefix string) map[string]string {
+	properties := make(map[string]string)
+	for _, logInfo := range logInfos {
+		prefix := []string{logInfo.podNameElement, "name", "log", logInfo.containerName}
```
paulosjca, line 47: I don´t think it needs "name" here.

I might add ...prefixes as a list of prefixes at the end of the arguments.

The "prefix" being caluclated in the function is not a prefix, it is the key.

I would just have key := strings.Join(append(prefix, logInfo.podNameElement, "log", logInfo.containerName), ".")

In fact, I might make this into a separate function, PodLogPropertyKey() with arguments of type LogInfo and ...prefix.

Then in the runner you can pass the prefix "pod" as an additional argument to put the properties in the right place.

tools/runner/properties.go
```diff
@@ -0,0 +1,74 @@
...
+	prefix := fmt.Sprintf("%s-", loadTestName)
+	podNameElement := strings.TrimPrefix(podName, prefix)
+	return podNameElement
+}
+
+// PodNameProperties creates pod property name to pod name map.
+func PodNameProperties(pods []*corev1.Pod, loadTestName string) map[string]string {
+	properties := make(map[string]string)
+	for _, pod := range pods {
+		prefix := []string{PodNameElement(pod.Name, loadTestName), "name"}
```
paulosjca, line 68: I would make PodNamePropertyKey() into a separate function with an extra argument ...prefix, in a similar way as with the log. Then I would pass the "pod" prefix in the runner.

tools/runner/logsaver.go
```diff
@@ -1,145 +1,103 @@
...
-// of the saved log.
-func (ls *LogSaver) SavePodLogs(ctx context.Context, loadTest *grpcv1.LoadTest, podLogDir string) (*SavedLogs, error) {
-	savedLogs := NewSavedLogs()
-
-	// Get pods for this test
-	pods, err := ls.getTestPods(ctx, loadTest)
-	if err != nil {
-		return savedLogs, err
-	}
+// SaveAllLogs saves all logs to files and return their LogInfo.
```
paulosjca, line 32: The godoc comments don't need an empty line, and don't need to repeat the name of the function. So if the first line is // SaveAllLogs saves..., the second line can just continue and explain: // This function ... We can fix the docstrings once the other issues are addressed.

tools/runner/logsaver.go
```diff
@@ -1,145 +1,103 @@
...
-func (ls *LogSaver) getPodLogBuffer(ctx context.Context, pod *corev1.Pod) (*bytes.Buffer, error) {
-	req := ls.podsGetter.Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{})
-	driverLogs, err := req.Stream(ctx)
+// SaveLog retrieves and save logs for a specific container.
+//
+// SaveLog retrieves logs from a container under given container
+// name within given pod,then save the logs to the given file
+// path, if there are logs to save. The function also returns
+// a point of the LogInfo object.
+func SaveLog(ctx context.Context, loadTest *grpcv1.LoadTest, pod *corev1.Pod, podsGetter corev1types.PodsGetter, containerName string, filePath string) (*LogInfo, error) {
```
paulosjca, line 71: If you made this function take the directory rather than the filepath, then you would be able to construct the filepath inside this function.

In fact, I would construct the logfilename in a separate function, from the pod name and the container name. Then inside SaveLog I would join that with the directory to construct the LogInfo.

paulosjca (approved):

PR #309, opened 2022-05-14
Add uniquifier to test case name.

wanlin31:
This commit add uniquifier field from annotation to the test case names.
This is needed since the PSM tests are running the same scenario with 
proxied and proxyless setups, using only scenario name for the test case
can not differentiate proxied and proxyless tests.

### DIFF
--- a/tools/cmd/runner/main.go
+++ b/tools/cmd/runner/main.go
@@ -96,7 +96,7 @@ func main() {
 	done := make(chan *runner.TestSuiteReporter)
 
 	for qName, configs := range configQueueMap {
-		testSuiteReporter := reporter.NewTestSuiteReporter(qName, logPrefixFmt, runner.TestCaseNameFromAnnotations("scenario"))
+		testSuiteReporter := reporter.NewTestSuiteReporter(qName, logPrefixFmt, runner.TestCaseNameFromAnnotations("scenario", "uniquifier"))
 		testSuiteReporter.SetStartTime(time.Now())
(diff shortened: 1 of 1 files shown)


tools/cmd/runner/main.go
```diff
@@ -96,7 +96,7 @@ func main() {
 	done := make(chan *runner.TestSuiteReporter)
 
 	for qName, configs := range configQueueMap {
-		testSuiteReporter := reporter.NewTestSuiteReporter(qName, logPrefixFmt, runner.TestCaseNameFromAnnotations("scenario"))
+		testSuiteReporter := reporter.NewTestSuiteReporter(qName, logPrefixFmt, runner.TestCaseNameFromAnnotations("scenario", "uniquifier"))
```
paulosjca, line 99: The uniquifier must not be used here, because it will prevent TestGrid from tracking successive runs.

What we should do instead is assign proxied and proxyless tests to different queues in the runner. Tests in each queue generate a separate xUnit report, so there will be no confusion. In order to do that we can add an annotation just for that and use the -annotation-key flag to ensure that this annotation is used instead of the "pool" annotation. I would suggest something like "queue: workers-8core-proxied" if we want to keep a similar naming scheme as for the non-PSM tests.

wanlin31: Thank you so much for pointing this out, I was not aware adding the uniquifier to the test case name would affect TestGrid, can you help me find a reference? Thank you.
I have tried the method you mentioned, worked great, I will close this PR since no change needed in the test-infra.

PR #315, opened 2022-06-29
Make collection of Prometheus data be controlled by an annotation.

wanlin31:
Currently we automatically collect Prometheus data for PSM
related tests, now some regular tests (same scenario as PSM
test) are also part of the PSM tests, we may want to collect
Prometheus data for those regular tests too. The current method
for checking if a test isPSM related is to check the existence of
xds-server and sidecar containers.

The default behavior is to not collect any Prometheus data. 
This PR adds an annotation `enablePrometheus` to 
elect to collect Prometheus data, once this annotation is
set to `true` in a LoadTest configuration, the 
Prometheus data would be collected for that particular test.

### DIFF
--- a/containers/runtime/driver/run.sh
+++ b/containers/runtime/driver/run.sh
@@ -35,7 +35,7 @@ if [ -n "${BQ_RESULT_TABLE}" ]; then
   fi
   if [ -r "${NODE_INFO_OUTPUT_FILE}" ]; then
     cp "${NODE_INFO_OUTPUT_FILE}" node_info.json
-    if [ -n "${SERVER_TARGET_OVERRIDE}" ]; then
+    if [ -n "${SERVER_TARGET_OVERRIDE}" ] || [ -n "${ENABLE_PROMETHEUS}" ]; then
       python3 /src/code/tools/run_tests/performance/prometheus.py \
         --url=http://prometheus.test-infra-system.svc.cluster.local:9090 \
         --pod_type=clients --container_name=main \
--- a/containers/runtime/driver/run.sh
+++ b/containers/runtime/driver/run.sh
@@ -35,11 +35,13 @@ if [ -n "${BQ_RESULT_TABLE}" ]; then
   fi
   if [ -r "${NODE_INFO_OUTPUT_FILE}" ]; then
     cp "${NODE_INFO_OUTPUT_FILE}" node_info.json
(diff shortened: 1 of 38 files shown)


paulosjca (commented):
I would also modify the PSM examples to add the `enablePrometheus` annotation, and modify the kokoro jobs to include it.

I also notice that there is no mention of Prometheus in the README files, we should update the documentation as well (perhaps as a separate PR).

containers/runtime/driver/run.sh
```diff
@@ -35,7 +35,7 @@ if [ -n "${BQ_RESULT_TABLE}" ]; then
   fi
   if [ -r "${NODE_INFO_OUTPUT_FILE}" ]; then
     cp "${NODE_INFO_OUTPUT_FILE}" node_info.json
-    if [ -n "${SERVER_TARGET_OVERRIDE}" ]; then
+    if  [ "$(dig +short -t srv prometheus.test-infra-system.svc.cluster.local)" ] && [ -z "${DISABLE_PROMETHEUS}" ]; then
```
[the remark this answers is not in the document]
paulosjca: On second thoughts, it may be better to test if SERVER_TARGET_OVERRIDE or ENABLE_PROMETHEUS are set. So Prometheus is used by default for PSM tests, and can be enabled with the annotation for other tests.

wanlin31:
The loadtest_examples.sh are updated in https://github.com/grpc/grpc/pull/30452

All examples from this PR are generated by  loadtest_examples.sh.

status/missing_pods_test.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+http://www.apache.org/licenses/LICENSE-2.0
```
paulosjca, line 8: I would leave the indentation as before.

paulosjca (commented):
This is fine, but you need to update the description to match.

status/missing_pods_test.go
wanlin31: Sorry I forgot you mentioned it should be 4 spaces instead of a tab. Updated.

The description is also updated, thanks for reminding me, almost forgot of it.

paulosjca (approved):

PR #317, opened 2022-08-02
Add PSM dashboard to OSS benchmark dashboard.

wanlin31:
This PR adds the PSM dashboard to the existing OSS dashboard.

PSM dashboard used the same Postgres-replicator and Grafana 
instances.


dashboard/Makefile
```diff
@@ -1,8 +1,11 @@
 GOCMD ?= go
 GOARGS = -trimpath
 CONFIG_TEMPLATE_DIR ?= config
+PSM ?= false
```
paulosjca, line 4: I would just make PSM ?= with nothing.

Then you can use conditions all the way like this:

ifeq ($(PSM),)
whatever you want
else
something else
endif

In particular, only set PSM_DASHBOARDS_CONFIG_DIR if PSM is set, and set REPLICATOR_CONFIG_TEMPLATE depending on whether PSM is set (no need for PSM version).

If you are feeling adventurous you can just add the PSM version of the dashboard location to the regular variable and then use the cp command directly with the two variables.

dashboard/README.md
```diff
@@ -40,29 +40,38 @@ dashboards with different configurations.
 [replicator]: cmd/postgres_replicator/README.md
 [replicatorconfig]: config/postgres_replicator/default/config.yaml
 
+## Deploy dashboard with PSM
```
paulosjca, line 43: The five lines above this one should be moved so they appear at the bottom of the section where they are used. There should be no spaces at the beginning of the line.

dashboard/README.md
```diff
@@ -40,29 +40,38 @@ dashboards with different configurations.
...
+
+PSM dashboard is not included in the deployment by default, following command is
```
paulosjca, line 45: I would not include this section.

Instead, I think we should update both the cloud_build.yaml and the manual build instructions to include the PSM variable in the table of variables, and explain how it works there.

dashboard/README.md
```diff
@@ -40,29 +40,38 @@ dashboards with different configurations.
...
+| --------------------------- | --------------------------------------- |
+| `BQ_PROJECT_ID`             | `grpc-testing`                          |
+| `CLOUD_SQL_INSTANCE`        | `grpc-testing:us-central1:grafana-data` |
+| `GCP_DATA_TRANSFER_SERVICE` | `postgres-replicator`                   |
+| `GCP_GRAFANA_SERVICE`       | `grafana`                               |
+| `GCP_PROJECT_ID`            | `grpc-testing`                          |
+| `GRAFANA_ADMIN_PASS`        | ...                                     |
+| `PG_DATABASE`               | `datasource`                            |
+| `PG_PASS`                   | ...                                     |
+| `PG_USER`                   | `grafana-user`                          |
```
paulosjca, line 67: I would add a line `PSM` | (see below).

Then below I would write "The dashboard for [PSM benchmarks](../index.md#psm-benchmarks) is only deployed if the `PSM` variable is set to a non-empty value, for instance `PSM=true`".

I would also update cloud_build.yaml to include the PSM variable, and then update the cloud build project to include `_PSM` with an appropriate value, for instance `true`.

wanlin31: I have updated the cloud build project to include _PSM as true.

paulosjca (approved):

PR #328, opened 2022-11-03
Ask driver to wait 20s before querying prometheus.

wanlin31:
This PR make the driver wait for 20s when we need to query Prometheus.
This would only affect the tests that need to query the Prometheus.


wanlin31:
This pr is closed since we decide to make the delay of query as a flag in https://github.com/grpc/grpc/blob/master/tools/run_tests/performance/prometheus.py

paulosjca:
> This pr is closed since we decide to make the delay of query as a flag in https://github.com/grpc/grpc/blob/master/tools/run_tests/performance/prometheus.py

You will still need to add the flag here, so you may want to reopen and update the PR once you are done adding the flag.

PR #332, opened 2022-11-15
Remove configuration that is not needed.

wanlin31:
This pr removes the redundant configuration for ServiceMonitor. 
We only query data for `job="kubernetes-cadvisor"` https://github.com/grpc/grpc/blob/8984a264b8c7dcc7c052e14e728d988eff0fb309/tools/run_tests/performance/prometheus.py#L83


paulosjca (approved):

wanlin31:
Tested with CI in psm-benchmarks-performance cluster, worked okay.

PR #336, opened 2022-11-19
Update go version to go1.17 everywhere

wanlin31:
This change matches https://github.com/grpc/grpc-go/pull/5477.

### DIFF
--- a/config/defaults_template.yaml
+++ b/config/defaults_template.yaml
@@ -27,7 +27,7 @@ languages:
   runImage: "{{ .RunImagePrefix }}cxx:{{ .Version }}"
 
 - language: go
-  buildImage: golang:1.16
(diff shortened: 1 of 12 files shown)


wanlin31:
This also affecting the CIs

paulosjca (commented):
I think you are missing changes in some places. I would run the following:

```shell
find . -type f | xargs -r sed -i -e s/golang:1.16/golang:1.17/g
go get ./...
go mod tidy -compat=1.17
```

I have already tested a complete CI run with these changes on the `benchmarks-dev1` cluster, so that should do the trick.

After submitting this, I would make the next release version 1.5.0, since we are changing compatibility.

paulosjca (approved):

PR #367, opened 2023-07-29
[Test] Write Prometheus stats to driver log.

paulosjca:
Write Prometheus stats to driver log, in addition to worker stats, even when not saving data to BigQuery.


paulosjca:
@ejona86

wanlin31 (approved):

PR #370, opened 2023-07-31
Add C++ experiments to PSM's replicator config

drfloob:
Due to the way the replicator is deployed, only the PSM replicator config is used. If the default is ever used, the PSM tables will stop getting populated.

@wanlin31 it may be worthwhile to unify the configs into the default replicator config, or separate them into multiple replicator jobs that are all running in parallel on disjoint sets of tables.

### DIFF
--- a/dashboard/config/postgres_replicator/psm/config.yaml
+++ b/dashboard/config/postgres_replicator/psm/config.yaml
@@ -18,3 +18,13 @@ transfer:
       dateField: metadata.created
     - name: psm_experimental_results_8core
       dateField: metadata.created
+  - name: e2e_benchmark_cxx_experiments
+    tables:
+    - name: results_32core_event_engine_listener
+      dateField: metadata.created
+    - name: results_8core_event_engine_listener
+      dateField: metadata.created
+    - name: results_32core_work_stealing
+      dateField: metadata.created
+    - name: results_8core_work_stealing
+      dateField: metadata.created


wanlin31 (approved):

paulosjca (approved):

wanlin31:
> Due to the way the replicator is deployed, only the PSM replicator config is used. If the default is ever used, the PSM tables will stop getting populated.

>  Yeah, that is expected if you only want to have a dashboard for regular benchmark data. Even the PSM table in PostgreSQL stopped being populated, the data would always be in BigQuery. Once someone want to create a dashboard for regular + PSM, AFAIR, the data would be copied to PostgreSQL from where it stopped.
 
> @wanlin31 it may be worthwhile to unify the configs into the default replicator config, or separate them into multiple replicator jobs that are all running in parallel on disjoint sets of tables.
If we unify the config, having PSM in the public dashboard would be our default, I don't see any issue if you think it is okay :)

paulosjca:
I don't think we should "unify the configs". I would add the configuration for experiments to the PSM dashboard config and ignore the other one. The reason why the other one exists is in case someone wants to deploy a dashboard and doesn't care about PSM: it is an example for others, rather than what we deploy ourselves.

drfloob:
> If we unify the config, having PSM in the public dashboard would be our default. And people who ever looking for deploy a personal dashboard would always have PSM tables in there, I think it might need further discussion :)

I'm just talking about the replicator configs, not the dashboard configs. IIUC, there's only one replicator. If you never deploy the replicator's `default` config because you always need PSM table replication (the trigger always deploys the PSM version right now), then you can merge those replicator configs together.

drfloob:
> The reason why the other one exists is in case someone wants to deploy a dashboard and doesn't care about PSM: it is an example for others, rather than what we deploy ourselves.

I see. Suggestion retracted, thanks for explaining.

PR #376, opened 2023-08-01
Update Node benchmark Dockerfile

murgatroid99:
This new Dockerfile only runs the benchmark for the pure JS implementation of the Node gRPC library. The other library has been deprecated for years and is not compatible with currently-supported versions of Node.


paulosjca (approved):

paulosjca:
This fixes the tests with prebuilt images, that are used in CI. Regular tests can be updated later.

PR #377, opened 2023-08-02
Update example tests.

paulosjca:
Tests are updated following https://github.com/grpc/grpc/pull/33968.

### DIFF
--- a/config/samples/cxx_example_loadtest.yaml
+++ b/config/samples/cxx_example_loadtest.yaml
@@ -99,7 +99,7 @@ spec:
             }
           }
         },
-        "warmup_seconds": 5,
+        "warmup_seconds": 30,
         "benchmark_seconds": 30
       }
     }
--- a/config/samples/node_purejs_example_loadtest.yaml
+++ b/config/samples/node_purejs_example_loadtest.yaml
@@ -5,12 +5,12 @@ apiVersion: e2etest.grpc.io/v1
 kind: LoadTest
 metadata:
   annotations:
-    scenario: node_to_node_generic_async_streaming_ping_pong_secure
+    scenario: node_purejs_to_node_generic_async_streaming_ping_pong_secure
     uniquifier: basic
   labels:
     language: node
     prefix: examples
-  name: examples-node-to-node-generic-async-streaming-ping-pong-secure-basic
+  name: examples-node-purejs-to-node-generic-async-streaming-ping-pong-secure-basic
 spec:
   clients:
   - build:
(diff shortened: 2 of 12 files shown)


drfloob (approved):
Are these auto-generated, or was this done manually?

murgatroid99:
If it matters functionally what the names say, then the new `node_purejs_to_node` strings should probably say `node_purejs_to_node_purejs`. And if it doesn't matter functionally, then it's probably not worth changing the names at all, and we can just say that we now use `node` to refer to the new pure JS implementation instead of the old wrapper implementation.

paulosjca:
They are generated by running the script updated by https://github.com/grpc/grpc/pull/33968.

https://github.com/grpc/grpc/blob/master/tools/run_tests/performance/README.md#generating-load-test-examples

murgatroid99:
So maybe that script should say `node` instead of `node_purejs`

paulosjca:
Currently it does matter functionally. This was likely a hack back in the day, to generate scenarios with two different configurations they were named as two different languages, `node` and `node_purejs`. This only affects the scenario naming and configuration, not the actual worker. If you want to drop the old `node` implementation completely and rename things so there is only `node`, we can do that as a follow-up step. The key to how this works is here:

https://github.com/grpc/grpc/blob/e0b51886be667c191a93885c98102872aa2b922f/tools/run_tests/performance/scenario_config.py#L1796

paulosjca:
Let's merge the examples since they are only examples and they show how the code is actually working now (for instance, in the experimental ci: https://github.com/grpc/grpc/pull/33963). We can update them again if that changes.

murgatroid99:
I think those scenarios might actually have run the pure JS client against the wrapper server, so the names were correct. I think they predate the existence of the pure JS server implementation.

But I wouldn't expect that we're using `scenario_config.py` anymore anyway, because it runs worker commands that don't correspond to the docker images we use in this repository.

paulosjca:
> So maybe that script should say `node` instead of `node_purejs`

If it does, then the scenarios will be generated with the old implementation. Feel free to deprecate the old implementation and make the change throughout.

murgatroid99:
The updated worker docker image only runs the pure JS implementation. It can't be configured to run the old implementation. So the scenario can describe itself as using the old implementation, but it won't actually run the old implementation.

paulosjca:
> The updated worker docker image only runs the pure JS implementation. It can't be configured to run the old implementation. So the scenario can describe itself as using the old implementation, but it won't actually run the old implementation.

This is a good reason to clean up `scenario_config.py`. We do generate the scenarios based on that. In any case, the test will be the same.

murgatroid99:
My perspective here is that there's no reason to keep the `purejs` string in the long run, so adding it to various places now just makes more work when we remove it later.

paulosjca:
Let's deprecate `node_purejs` instead.

paulosjca:
Superseded by https://github.com/grpc/test-infra/pull/378.

PR #382, opened 2023-09-14
Explicitly pass through the name of the Prometheus result file.

wanlin31:
This change explicitly provide the name of the Prometheus result file.

### DIFF
--- a/containers/runtime/driver/run.sh
+++ b/containers/runtime/driver/run.sh
@@ -35,7 +35,7 @@ if [ -n "${SERVER_TARGET_OVERRIDE}" ] || [ -n "${ENABLE_PROMETHEUS}" ]; then
       --url=http://prometheus.test-infra-system.svc.cluster.local:9090 \
       --pod_type=clients --container_name=main \
       --container_name=sidecar --delay_seconds=20 \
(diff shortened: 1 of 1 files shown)


containers/runtime/driver/run.sh
```diff
@@ -35,7 +35,7 @@ if [ -n "${SERVER_TARGET_OVERRIDE}" ] || [ -n "${ENABLE_PROMETHEUS}" ]; then
       --url=http://prometheus.test-infra-system.svc.cluster.local:9090 \
       --pod_type=clients --container_name=main \
       --container_name=sidecar --delay_seconds=20 \
-      --export_file_name=prometheus_query_results.json
+      --export_file_name=prometheus_query_result.json
```
paulosjca, line 38: Just to make it a little bit less error-prone, I would define a constant before the if statement, something like this:

```shell
declare -r PROMETHEUS_QUERY_RESULT_FILE=prometheus_query_result.json
```

and then use the constant in the two places where it is needed.

wanlin31: Done.

paulosjca (approved):

PR #384, opened 2023-09-15
Fix checkout for driver build.

paulosjca:
This is a workaround for an error when checking out gRPC v1.58.1 for the driver build.

### DIFF
--- a/containers/runtime/driver/Dockerfile
+++ b/containers/runtime/driver/Dockerfile
@@ -32,7 +32,7 @@ WORKDIR /src/code
 
 RUN git clone https://github.com/$REPOSITORY.git .
 RUN git submodule update --init
-RUN git checkout $GITREF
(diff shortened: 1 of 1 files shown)


paulosjca:
There is a submodule that causes a problem when cloning and then checkout out `v1.58.1`. Error message:

```shell
error: The following untracked working tree files would be overwritten by checkout:
        third_party/upb/.bazelci/presubmit.yml
        [...other files under third_party/upb...]
Aborting
```

wanlin31 (approved):
Seems like we are forcing the change of the working tree:
```
       -f, --force
           When switching branches, proceed even if the index or the working tree differs from HEAD, and even if there are untracked files in the way.
           This is used to throw away local changes and any untracked files or directories that are in the way.
``` 
Can I ask we we need to force it?

PR #385, opened 2023-09-15
Update base images used for tests to golang:1.20.

paulosjca:
grpc-go requires go1.19. Infrastructure will be updated separately.


wanlindu (commented):
Do we need to update go.mod for this change?

paulosjca:
> Do we need to update go.mod for this change?

No, this is not changing anything in the infrastructure, only in the base images used for tests. The go.mod specifies the minimal go version *required* to build the **infrastructure**, this change updates the go version *actually* used to build **grpc-go**.

wanlindu (approved):

PR #390, opened 2024-01-18
Use driver from gRPC core 1.60.0

lepistone:
This is in preparation for using a driver that contains the fix grpc/grpc#35384.

There are no releases yet that contain the fix, but at least we can start by getting the latest and fixing the related problems.

There is in fact a pre-release with the fixed driver: [v1.61.0-pre1](https://github.com/grpc/grpc/releases/tag/v1.61.0-pre1).

Once this fix is merged, the fixed driver can be built by setting `DRIVER_VERSION` to `v1.61.0-pre1` or `master`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 TEST_INFRA_VERSION ?= latest
 
 # Version of gRPC core used for the gRPC driver
-DRIVER_VERSION ?= v1.58.1
+DRIVER_VERSION ?= v1.60.0
 
 # Prefix for all images used as clone and ready containers, enabling use with
 # registries other than Docker Hub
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -2,7 +2,7 @@
 TEST_INFRA_VERSION ?= latest
 
 # Version of gRPC core used for the gRPC driver
-DRIVER_VERSION ?= v1.58.1
+DRIVER_VERSION ?= v1.60.0
```
paulosjca, line 5: Cound you change this version to `v1.62.1`? Thanks!

lepistone: Done, thank you!

paulosjca (approved):

PR #392, opened 2024-03-11
Make it possible to configure a metrics port in pods

fdfzcq:
This PR currently includes changes from https://github.com/grpc/test-infra/pull/390

We (Spotify) are currently using this framework to set up automated benchmarking for our internal gRPC. While we have metrics from the driver for performance comparison, we would like to scrape and see actual metrics from the client/server in many cases. This PR adds a configuration `metricsPort` for both client and server. A metrics port with configured port number is exposed in the pod if configured, otherwise no extra ports are exposed.

### DIFF
--- a/podbuilder/podbuilder_test.go
+++ b/podbuilder/podbuilder_test.go
@@ -388,7 +388,7 @@ var _ = Describe("PodBuilder", func() {
 				Expect(getNames(runContainer.Ports)).NotTo(ContainElement("metrics"))
 			})
 
-			It("does not expose the metrics port if not set", func() {
+			It("exposes the metrics port if set", func() {
 				client.Run = []corev1.Container{{}}
 				client.Run[0].Name = config.RunContainerName
 				client.Run[0].Command = []string{"go"}
--- a/podbuilder/podbuilder_test.go
+++ b/podbuilder/podbuilder_test.go
@@ -374,6 +374,36 @@ var _ = Describe("PodBuilder", func() {
 				Expect(getValue("driver", "ContainerPort", runContainer.Ports)).To(BeEquivalentTo(config.DriverPort))
 			})
 
+			It("does not expose the metrics port if not set", func() {
+				client.Run = []corev1.Container{{}}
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -2,7 +2,7 @@
 TEST_INFRA_VERSION ?= latest
 
 # Version of gRPC core used for the gRPC driver
-DRIVER_VERSION ?= v1.58.1
+DRIVER_VERSION ?= v1.60.0
```
paulosjca, line 5: Could you change this version to `v1.62.1`? Thanks!

paulosjca (changes requested):

Makefile
lepistone: @fdfzcq @paulosjca  FYI I made the same change in #390 that this PR is based on. I suppose the easiest is to get #390 merged first, then rebase this one. In any case it's an easy conflict to resolve.

fdfzcq: I removed the commits from @lepistone in this PR now

podbuilder/podbuilder_test.go
```diff
@@ -374,6 +374,36 @@ var _ = Describe("PodBuilder", func() {
...
+
+				pod, err := builder.PodForClient(client)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(pod.Spec.Containers).ToNot(BeEmpty())
+
+				runContainer := kubehelpers.ContainerForName(config.RunContainerName, pod.Spec.Containers)
+				Expect(getNames(runContainer.Ports)).NotTo(ContainElement("metrics"))
+			})
+
+			It("does not expose the metrics port if not set", func() {
```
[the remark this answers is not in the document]
paulosjca: Fixed in next commit.

paulosjca (approved):

podbuilder/podbuilder_test.go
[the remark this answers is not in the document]
fdfzcq: oops sorry I missed this comment, thank you for fixing it! @paulosjca

PR #399, opened 2024-11-22
Update golang and driver versions.

paulosjca:
- Updates the go version for tests that do not use prebuilt images to go 1.23.
- Updates the go version for github actions to 1.21.x (build currently fails with go version >= 1.22.0).
- Updates the go version for the controller build to 1.21.
- Updates the driver version to (grpc) 1.68.0.

### DIFF
--- a/config/defaults_template.yaml
+++ b/config/defaults_template.yaml
@@ -27,7 +27,7 @@ languages:
   runImage: "{{ .RunImagePrefix }}cxx:{{ .Version }}"
 
 - language: go
-  buildImage: golang:1.22
+  buildImage: golang:1.23
   runImage: "{{ .RunImagePrefix }}go:{{ .Version }}"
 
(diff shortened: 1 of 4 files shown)


config/defaults_template.yaml
```diff
@@ -27,7 +27,7 @@ languages:
   runImage: "{{ .RunImagePrefix }}cxx:{{ .Version }}"
 
 - language: go
-  buildImage: golang:1.20
+  buildImage: golang:1.22
```
arjan-bal, line 30: We can bump it to 1.23 so that we don't don't need to do this again when grpc-go drops support for 1.22.

arjan-bal: Or we could use `golang:latest`.

paulosjca: Done.

PR #417, opened 2026-06-15
Update Bazel base image for cxx worker to 8.7.0

ganesh-tn:
The bazel image v8.7.0 was updated in the Grpc repository last week (https://github.com/grpc/grpc/pull/42392). We are moving to this newer version because the previous 8.0.1 release caused compatibility issues.


paulosjca (changes requested):
This is for more than cxx. Can you update the description to include the changes athata are made? It looks like the commit is also named after 8.0.1, although that is less important.

PR #421, opened 2026-07-08
Fix Python pre-built worker startup crash

ganesh-tn:
Fix Python pre-built worker startup crash

### DIFF
--- a/containers/pre_built_workers/python/Dockerfile
+++ b/containers/pre_built_workers/python/Dockerfile
@@ -32,6 +32,10 @@ RUN git clone https://github.com/$REPOSITORY.git .
 # checkout, but skip updating the submodules since they're not use by bazel build
 RUN git checkout $GITREF
 
+# Patch python files to fix shadowing of 'src' package by other Bazel dependencies (like typing_extensions)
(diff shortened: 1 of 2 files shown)


containers/pre_built_workers/python/Dockerfile
```diff
@@ -32,6 +32,10 @@ RUN git clone https://github.com/$REPOSITORY.git .
 # checkout, but skip updating the submodules since they're not use by bazel build
 RUN git checkout $GITREF
 
+# Patch python files to fix shadowing of 'src' package by other Bazel dependencies (like typing_extensions)
```
paulosjca: Perhaps this can be fixed by changing PYTHONPATH, rather than editing the files.

The same should be applied to the non-prebuilt Dockerfile too.

See an example implementation here: https://github.com/paulosjca/test-infra/commit/72c5a7236a34cdde38dbc056327360561d947eec

paulosjca (changes requested):

containers/pre_built_workers/python/Dockerfile
ganesh-tn: Simplified fix via PYTHONPATH

paulosjca (approved):

PR #424, opened 2026-07-28
Fix build for java, cxx, init clone and driver images

arjan-bal:
This PR fixes broken docker image builds in preparation for a minor release for adding gRPC Rust support (https://github.com/grpc/test-infra/pull/423).

## Tested
Verified that on-the-fly tests pass for CPP, Java with the new driver image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 TEST_INFRA_VERSION ?= latest
 
 # Version of gRPC core used for the gRPC driver
-DRIVER_VERSION ?= v1.68.0
+DRIVER_VERSION ?= v1.82.1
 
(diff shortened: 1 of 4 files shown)


arjan-bal:
> LGTM, for CPP and common changes. Please take one for java change

The java run image is copied from the pre-built worker Dockerfile that was updated earlier in https://github.com/grpc/test-infra/pull/408: https://github.com/grpc/test-infra/blob/db69bd3787aab8c03f80402016eed77cec3959a8/containers/pre_built_workers/java/Dockerfile#L47-L58

I've also verified that the image runs the server and client and the benchmarks reports the expected QPS.

